PR #138, opened 2015-01-06
include where clause checking if a field is empty

dpfavand:
This issue came up on the irc today - 
nmarques attempted to use {% setcontent members = 'staff' where {position: "!''"} %} to return records where staff was not empty, but found it didn't work. Instead, nmarques figured out that '!' worked: {% setcontent members = 'staff' where {position: '!'} %} 

Since this seems to me to be an unexpected behavior, I figured it should be documented. (Alternatively if it's a bug, then an issue might be opened?)


bobdenotter:
I have never used this before, but I think it's a clever trick. Now that it's documented, it's definitely not a bug. ;-)

Thanks, merging in!

PR #139, opened 2015-01-13
Remove timthumb rewrite rule from htaccess

nlemoine:
Related to https://github.com/bolt/bolt/issues/2451


bobdenotter:
Good catch! Thanks!

PR #150, opened 2015-02-13
Added some details to 'isSafe()' description.

bobdenotter:


GawainLynch:
The only thing I want to know about this is, how on earth did I add a trailing space on a line!

PR #153, opened 2015-02-23
Chmod cache notification

evertalbers:
I am not sure if Git installation also gives these notifications, since I've never tried that. And the text could be a little more humouristic because it's not an actual problem, but I didn't have any inspiration.


bobdenotter:
Thanks! :+1:

PR #160, opened 2015-03-05
added .woff2 access permission into nginx configuration

Leon0824:
Bolt back-end loads Source Sans with woff2 extension.


GawainLynch:
Good catch, thanks!

PR #161, opened 2015-03-08
If we redirect, we should stop doing other stuff..

bobdenotter:


rarila:
I wonder if it should be be ``exit`` instead of ``die()``

PR #164, opened 2015-03-08
Added the latest options for `select` fields.

bobdenotter:


source/howto/select-field-type.md
```diff
@@ -160,5 +160,26 @@ comma. The following example will display the 'first name' and 'last name' of ea
 ```
 
 
+If the list is growing longer, there are a few ways to make it more manageable:
 
+```apache
```
GawainLynch: Shouldn't these be `yaml`

GawainLynch:
No real objections to this, but just flagging it. Would it be better to do images and bits in English?

source/howto/select-field-type.md
bobdenotter: highlight.js doesn't support `yml`, and `apache` looks nicer than whatever it picks as a fallback. ;-)

GawainLynch: Ah, highlight! Oh, look, :koala:

bobdenotter:
Yeah, i was lazy. Had that example lying around. Think it's an issue? If so, i'll make an example screenshot in english.

GawainLynch:
Nah, just flagging it. Someone's CDO will kick in a some point anyway.

PR #173, opened 2015-03-30
Update install config for Nginx

### LINKED ISSUE #170: More secure Nginx config
As originally raised by @BrandonHinkel in his article [Securing BoltCM on Nginx](https://brandonhinkel.com/entry/securing-boltcm-on-nginx)

```nginx
location ~* (?:/files/|/app/database/|/app/cache/|/app/config/|/theme/|/extensions/?)(.*)\.php$ {
    return 403;
}
```
Mostly here so I don't forget to follow up, as I already did.

~~Extensions can have their own local assets, so need to look closer at that.~~

**EDIT:** Scratch that last comment, he's specifying PHP files.

GawainLynch:
* Simplify the YAML/Twig block
* Add some comments
* Don't allow PHP files to be called from upload, app, theme and extensions directories (fixes #170)

### DIFF
--- a/source/installation.md
+++ b/source/installation.md
@@ -338,8 +338,17 @@ server {
         add_header Cache-Control "public, mustrevalidate, proxy-revalidate";
     }
 
-    location = /robots.txt { access_log off; log_not_found off; }
-    location = /favicon.ico { access_log off; log_not_found off; }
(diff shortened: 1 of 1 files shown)


bobdenotter:
Looks good, merging in!

PR #175, opened 2015-03-31
Note about conflicting PHP modules

### LINKED ISSUE #174: Requirements page / Disable Zend Guard Loader
https://docs.bolt.cm/requirements

Add:
Zend Guard Loader must be disabled

I almost gave up on Bolt because I wasn't able to get it installed on my VPS. After posting a question on Stack Overflow with no response after almost 2 months (http://stackoverflow.com/questions/28359152), then some trial and error after comparing with one of my other VPSes, I found out that Zend Guard Loader must be disabled. Works good now.

GawainLynch:
Fixes #174

### DIFF
--- a/source/requirements.md
+++ b/source/requirements.md
@@ -24,6 +24,10 @@ are default settings, and Bolt should work out-of-the-box.
   - The cUrl extension
   - The GD Extension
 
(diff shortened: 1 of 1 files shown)


bobdenotter:
Ha! :-)

PR #186, opened 2015-04-14
Add 'where-goes-what' section to the installation instructions.

evertalbers:
You may also want to cross-link the items to the pages where they are explained more in detail.

I was also wondering if it is a good idea to add something like "After changing this file, reload the website to see the effect (or for contenttypes: update the database)".


bobdenotter:
Seems like a good idea to add this. Merging this in for now. Crosslinks and expanding can be done at a later time .

PR #191, opened 2015-04-21
Assign return value from app.config.set() to a variable so it's not output to the browser

GawainLynch:
Fixes https://github.com/bolt/bolt/issues/3414


CarsonF:
Wait why was `set ret =` included?

GawainLynch:
> Wait why was `set ret =` included?

It b0rks… I gave up caring past that.

CarsonF:
I knew it! I didn't think it was that easy. Oh well this is a workaround anyways

GawainLynch:
At least I am testing my changes… to docs… 'cause… yeah…

PR #197, opened 2015-05-03
Local extensions do appear in Extend page

frankbirzle:
After following the steps of the How To document 'Installing Local Extensions' i _could_ actually see my extension in the Extend page.

I removed the statement that local extensions do not appear in the Extend page and added an additional sentence in Step 3 stating the opposite.


GawainLynch:
Ah cool, yeah I forgot to update the docs when I finally fixed that in core!

PR #208, opened 2015-05-25
Document 503 status code

tvlooy:
I had to look this up / was missing this information in the docs. It's might be a bit too technical for the docs, but the fact that I checked the docs before trying the feature could mean that other people are also missing this information.


bobdenotter:
Thanks, Tom!

PR #209, opened 2015-05-25
Explain markdown and how to serve the docs

tvlooy:
Some people might not know markdown, some people might think they need to built the documentation or think it's not possible to just serve it.

I had to figure this out, so I think this information belongs in the docs.


bobdenotter:
:+1:

PR #210, opened 2015-05-25
Reset password options

tvlooy:
Not sure if this (howto) is the right place to put this docs.

The user:reset-password command is an open PR bolt/bolt#3560, this PR should not be merged until the code is merged.


GawainLynch:
This works for me, but @bobdenotter is far better at understanding what normal people want… I shall defer to him :grin:

bobdenotter:
I'm fine with this.. Perhaps we should add the 'as a last resort, truncate the bolt_users table' thing?

PR #213, opened 2015-05-30
add tip to autload local extension

tonicospinelli:
after some minutes to discover why local extensions not been autoloaded... 

why the composer not load my local extension classes? I created the composer file and execute the `composer dump-autload`, but why doesn't load?

So, :boom: I had an idea, I did add the vendor/autload.php in my init.php file from my local extension and it solved my problem.

It works good for me and it can help many people which have local extension!!

If I forgot to change others places, tell me, please!!

### DIFF
--- a/source/extensions/config.md
+++ b/source/extensions/config.md
@@ -58,8 +58,7 @@ and namespaces. There are three files you need to edit, `composer.json`,
 The above steps will get you started, and below is some more indepth information
 about the configuration.
 
-<p class="note"><strong>Tip:</strong> As noted in <a href="/howto/installing-local-extensions">Installing Local Extensions</a>, local extensions have no autoloader by default. Use <code>include_once</code> in Extension.php to load any additional class files you may create. If you move the extension to an external repository, remove the <code>include_once</code> lines.</p>
-
+<p class="note"><strong>Tip:</strong> As noted in <a href="/howto/installing-local-extensions">Installing Local Extensions</a>, local extensions need some caution.</p>
 
 ### The JSON file
 
--- a/source/howto/installing-local-extensions.md
+++ b/source/howto/installing-local-extensions.md
@@ -6,10 +6,11 @@ extensions site.
 
 There are a couple of caveats:
 
-  - There is no autoloader by default
-  - Must be located in `{web_root}/extensions/local/{author_name}/{extension_name}/`
+  - Autoloading is NOT guaranteed unless you have a properly defined composer.json file
+  - MUST be located in `{web_root}/extensions/local/{author_name}/{extension_name}/`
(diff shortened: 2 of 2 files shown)


GawainLynch:
Hi again, @tonicospinelli 

OK, so I was thinking about this in the shower yesterday… it has annoyed a lot of us for a while that we can't use an autoloader for Bolt extensions… then it hit me that the way I had been approaching the problem was the wrong way, better still there is an easy workaround!

So there is now master https://github.com/bolt/bolt/pull/3606 and soon 2.2 covered https://github.com/bolt/bolt/pull/3607

Given that you put in the effort here, and it is very much appreciated, would you like to update this PR to add details of how to do it now? (and get your name on the board for doing it :grin:)

tonicospinelli:
Wooww... @GawainLynch on fire! Because that I love this CMS, the team is amazing!!

Nice job!! I'll update this and call you :+1:

tonicospinelli:
@GawainLynch 
I did read the docs about local extension and I came to a conclusion, that not make sense keeping the doc `howto/installing-local-extensions.md` because the steps to create a local extension is the same as "published" extension.

So, I going to remove this page and its references, does it make sense for you?

GawainLynch:
That doesn't make sense. There is a very distinct differences between local and published extensions. 
* Location and paths are very specifically required
* Autoloading is not guaranteed unless you have a properly defined composer.json file
* Updates are not available though the web UI.

That page should absolutely stay, and be updated. :grin:

tonicospinelli:
Humm, those points are relevant, but I think they are only notes about local extension on extensions/config.md and no need a page.

GawainLynch:
I had a chat with @bobdenotter about this on IRC (please do feel free to join us there any time) and we both agree that they are different and should be on separate pages.

The idea for "local" extensions was more focused on backend developers/administrators so they didn't have to worry about git and Satis for small extensions, and it's not something we see fitting with [The Bolt Manifesto](https://docs.bolt.cm/manifesto), specifically:
* Bolt aims at three groups of users: Editors (content writers and editorial staff), Frontend developers (HTML/CSS specialists and implementers) and Backend developers (coders, devops and sysops). All three are equally important.

For people like you and I, local extensions are really powerful and a great thing to have, we just don't think that confusing "normal" users is the best way to go.

tonicospinelli:
Don't worry, I understand you. So I going to avoid my last commits and send a new commit with changes in note from extension config and in local extension page :wink:

GawainLynch:
Awesome :+1:

tonicospinelli:
@GawainLynch can you review my last commit? :grin:

GawainLynch:
Awesome mate! It's just after midnight here now, so I'll look closer in the morning and merge for you!

PR #216, opened 2015-06-09
Some fixes and tweaks.

bobdenotter:


GawainLynch:
Can you check this against the work done in #204

bobdenotter:
It was duplicating a lot of things, only yours is more thorough (and for 2.3). I've updated the PR to only _add_ a small section, which should give no conflicts with #204.

GawainLynch:
Awesome-sauce

PR #226, opened 2015-06-26
Updated StorageEvent(s) classes and constants

sbani:
Classes changed, so we need to update them in docs too.


GawainLynch:
:+1:

PR #227, opened 2015-07-08
Added update information for local extensions

sbani:
Give the info how to import/update a local extension.
Add the info that you need to run composer update to create new autoloader file.


GawainLynch:
:+1:

PR #235, opened 2015-08-09
Clarfified handling of config_local.yml

royallthefourth:
The current documentation for config_local.yml seemed a little ambiguous. I added a sentence to clarify that Bolt doesn't have an environment setting and that it's up to the deployment process to keep config_local.yml off the server.


GawainLynch:
Good stuff… Thank you!

PR #241, opened 2015-08-26
Really minor tweaks.

bobdenotter:


GawainLynch:
Dot points and headings should never be terminated with a period

It is like finishing a sentence without a period

There are some things that

PR #252, opened 2015-10-03
Add a practical example of an hourly drop down using the select field

geoff-winner:
@bobdenotter,

I added this example to the documentation!


GawainLynch:
:+1:

PR #253, opened 2015-10-06
fixed config filename and add explanation where to find the real config file

PhillippOhlandt:
That has confused me as I started with my first extension.


GawainLynch:
Cool! Thanks mate!

PR #254, opened 2015-10-07
Setting Up cURL SSL/TLS Certificate Authority Certificates

GawainLynch:
This is based on this [comment](https://github.com/bolt/bolt/issues/3575#issuecomment-146018366) by @dadaxr


bobdenotter:
Excellent! I saw that comment, and thought "Perhaps we should add this to the docs" :-)

GawainLynch:
> Excellent! I saw that comment, and thought "Perhaps we should add this to the docs" :-)

Thinking along the same lines, as per our usual level of Team Awesomeness :-)

PR #262, opened 2015-10-18
Change docs favicon per #245

### LINKED ISSUE #245: Feature request: Give docs site favicon that is different from Bolt back-end favicon
I often have the docs site open in one browser tab and the backend of a site that I'm building in another. Both tabs then show the same favicon. "Confusing" is too big a word, but I could dedicate 4 more braincells to the actual development if both favicons were different. 

Just a different colour would be enough for me, but maybe colourblind users would prefer a different form.

SahAssar:
Credit to @evertalbers for the icon!

Closes #245


GawainLynch:
Transport of the discussion on IRC to here for history/record…
```
<Bopp> must've missed that :-)
<Bopp> yeah, no.. 
<gawainlynch> Yeah… simmering for a while
<Bopp> I don't like the icon. 
<SahAssar> For me that is just a pr version of the issue. I have no real "no dog 
in this hunt".
<Bopp> its too, euhm, red
<Bopp> is evertalbers here?
<SahAssar> Although I do think it would be nice to have different favicons for the 
docs, cheatsheet and favicon. If it's the color that should change is a different 
question though.
<Bopp> yes, this.
<gawainlynch> Yeah… it gets confusing here
<Bopp> Especially how i want to get rid of the orange/red over time
<evertalbers> BoppPad: I just read the discussion about the favicon for the docs. 
Actually, I don’t really care what it looks like, as long as it’s different fom the 
default favicaon. At the time when I made it there was already a discussion 
going on about the logo etc. and I did not want to add more design opinions, 
so I only changed the colour.
<evertalbers> Accessibility-wise, it would even be better to make the form different 
from the default icon, for the colour-blind among us.
```

evertalbers:
Actually, I don’t care much about what it looks like, as long as it’s different fom the default favicon. At the time when I made it there was already a discussion going on about the logo etc. and I did not want to add more design opinions, so I limited myself to changing the colour.
Accessibility-wise, it would even be better to make the form different from the default icon, for the colour-blind among us.

bobdenotter:
Spring cleanup! Lets think about the set of icons, when we're going to redo the backend.

PR #267, opened 2015-10-29
New navigation bar design

anketwokings:
New navigation:
- Cleaner looking list
- Colors are equal to the new Bolt backend colors.


GawainLynch:
@bobdenotter says :+1: and I don't have a penguin to throw at you… so let's just merge instead :stuck_out_tongue_winking_eye:

rarila:
Oh, I don't think I like this one: [``<a>{{ name }}</a>``](https://github.com/bolt/docs/pull/267/files#diff-4a7d4785e1bfa5e47763b33b71c236abR6)

PR #272, opened 2015-10-30
Small fixes to new sidebar

anketwokings:
I fixed the lineheight and the padding, so the submenus look better. Plus, re-added the `<a>` because of a11y.


GawainLynch:
:+1:

PR #273, opened 2015-11-02
Add bolt installer instructions

PhillippOhlandt:
Bob said it's ok so I only accept change requests for the wording :P

### DIFF
--- a/source/installation.md
+++ b/source/installation.md
@@ -2,15 +2,14 @@ Installing Bolt
 ===============
 
 This page explains the various ways of installing Bolt. You can either use the
-command-line or your FTP-client to install it. There are four ways to install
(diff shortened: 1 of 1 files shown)


PhillippOhlandt:
Moved the instructions, maybe this is a better place? Feedback on wording?

GawainLynch:
As of our last discussion, neither @bobdenotter nor I are keen on this approach generally.There is some validation towards having a separate page of "idea/options/etc" to include this as an option. 

That said, I am very concerned that your installer could become a broader problem for all of us…:
> But, I read it in your official documentation and it doesn't work!!!!!1!!1!! You sux!!11!

:grin: 

Also I am rather, um, worried about that code and the approach used as well… Not to mention these things should be core projects if we're going to support them in this way.

PR #277, opened 2015-12-10
Prevent overwriting .htaccess file

evertalbers:
Inspred by https://discuss.bolt.cm/d/123-updating-using-ssh-overwriting-custom-htaccess

### DIFF
--- a/source_docs/updating.md
+++ b/source_docs/updating.md
@@ -38,7 +38,7 @@ curl -O http://bolt.cm/distribution/bolt-latest.tar.gz
 tar -xzf bolt-latest.tar.gz --strip-components=1
 chmod -R 777 files/ app/database/ app/cache/ app/config/ theme/ extensions/
 ```
-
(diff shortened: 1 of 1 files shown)


SahAssar:
Might just be my CDO acting up, but could you wrap the text at 80 chars?

evertalbers:
If only I knew how to do that with the GitHub web client…

Sorry for that.

SahAssar:
@evertalbers: Here, copy-paste this:
```
<p class="note"><strong>Note:</strong> If you have made edits to your .htaccess
file, make a backup of this file on your own computer (e.g. renamed
clientname_v01.htaccess so it becomes visible) or make sure that it does not
get overwritten by changing the permissions.</p>
```

evertalbers:
Just made a new pull request with that. 
Probably not the way to go, but I hope it works.

GawainLynch:
Line lengths we can fix on merge, but renaming `.htaccess` to a "visible" name is a minor security problem, as that then becomes publicly accessible and potentially disclosing server configuration information.

I might make an edit after merge, prior to pushing to the server.

evertalbers:
Thanks!

PR #282, opened 2015-12-20
Add JSON to list of denied file types for Nginx

GawainLynch:


bobdenotter:
Should we block JSON by default? We often make sites with maps, where the data is pulled in as JSON. For a recent example, see: http://vuurwerkoverlast.nl/page/kaart, where we pull in http://vuurwerkoverlast.nl/files/static/meldingen.json

GawainLynch:
Well… [you tell me](http://vuurwerkoverlast.nl/extensions/composer.json) :wink:

GawainLynch:
But that doesn't mean that people can't turn it on as needed…

bobdenotter:
right.. Well, changing it in the default `.htaccess` like this would mean breakage for a lot of people, without a clear indication of what/how/why.. 

For example, if it's merged into Bolt 2.2.15, sites like itgwo.nl and 200jaarstatengeneraal.nl would just stop working. 

I'd rather we specifically blacklist `package.json`, `composer.json` and `bower.json`.

GawainLynch:
Do they run Nginx?

GawainLynch:
Oh I see… comments are not matching PRs… Yeah, I have no problem updating https://github.com/bolt/bolt/pull/4610 to limit that. But this PR is only what is documented and that won't break existing sites.

bobdenotter:
Ah, yes. I lumped them together in my head. :-)

I think the "default nginx config" as documented should match the .htaccess one functionality-wise. So, imho it'd be best to update this one accordingly as well.

GawainLynch:
Updated and tested locally.

bobdenotter:
Looks good to me!

GawainLynch:
Teamwork FTW!

PR #295, opened 2016-01-31
Make 'BuildDocumentation' work recursively on folders.

bobdenotter:


bobdenotter:
Before:

screen_shot_2016-01-31_at_13_30_27

After:

screen_shot_2016-01-31_at_13_50_39

src/Command/BuildDocumentation.php
```diff
@@ -33,32 +33,55 @@ protected function configure()
...
 
+        foreach ($input->getArgument('branches') as $branch) {
+            $this->getSubTree($branch, $output, './source_docs/');
+            $versionsArray[$branch] = $foldername;
+        }
+
+        $dumper = new Dumper();
+
+        // Write the passed in versions to the yml file
+        $path = getcwd() . '/app/' . $this->versions;
```
GawainLynch: What if the binary is moved to another directory?

bobdenotter: The binary of what? This is just the docs site, so we can just choose not to do so, right?

GawainLynch: This is a `Command` class so it "creates" a "binary", or more explicitly something executable

GawainLynch: Also, technically it can, and should be able to, be called from anywhere

PR #296, opened 2016-01-31
Add a separate link prefix so assets and hrefs can use different variables

rossriley:


GawainLynch:
I have a merge conflict resolution half done locally… WRT `src/Command/BuildDocumentation.php` I presume we don't need the changes from #295 for that file @bobdenotter ?

bobdenotter:
I don't think this is needed. It'll look quite different soon: https://github.com/bolt/docs/tree/refactor-codebase

GawainLynch:
> I don't think this is needed. 

This PR you mean?

bobdenotter:
Yes.. The changes in index.php are not relevant after the refactor. And i've simplified the recursion in my version of the code, as pointed out by Ross. See: https://github.com/bolt/docs/blob/master/src/Command/BuildDocumentation.php#L59 

(which I accidentally pushed to master instead of to my branch, oops)

GawainLynch:
OK then… Second BPFL act for the day…

PR #308, opened 2016-02-09
Update contenttypes-and-records.md

xpika:
fixed grammatical error


GawainLynch:
Looks like the a good catch :wink: 

:+1:

PR #311, opened 2016-02-15
remove misleading line about more information

jadwigo:
because information about installing bolt is not what you're looking for at this point


GawainLynch:
Cherry-picked into master ;-)

PR #314, opened 2016-03-08
add docs for dynamic replacement feature

rossriley:
Covers bolt/bolt#3005


bobdenotter:
So, _that_ is what it does! ;-)

PR #330, opened 2016-03-25
Algolia search

PhillippOhlandt:
Styles should work for now, otherwise :hamster: can change them like he wants :)


GawainLynch:
:+1:

PR #345, opened 2016-04-04
enable versioned search results

PhillippOhlandt:
Assuming the `version` variable in twig holds `master` or `release-2-2` depending on which version you are.


bobdenotter:
Thanks! :+1:

PR #346, opened 2016-04-04
Fixing paths to images.

### LINKED ISSUE #342: Images are broken
I just saw that the images are broken, because the file path gets redirected to the versioned path too.

Look here:
https://docs.bolt.cm/release-2-2/howto/making-sure-htaccess-works

bobdenotter:
Fixes #342


PhillippOhlandt:
whoop, looks good to me (I think)

PR #348, opened 2016-04-07
Remove duplicated page

PhillippOhlandt:
I can't find the markdown file for that page in the source. Am I blind or is it not there? Same for the `release-process` file (which doesn't need to be deleted but isn't there too.).

### DIFF
--- a/menu_docs.yml
+++ b/menu_docs.yml
@@ -127,4 +127,3 @@ other:
     credits: "Credits"
     roadmap: "Roadmap"
     release-process: "Release process"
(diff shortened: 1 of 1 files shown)


bobdenotter:
Looks like something that got mixed up, when restructuring docs. :-)

bobdenotter:
> Same for the release-process file

I'll ninja-fix that!

PhillippOhlandt:
Yeah it's weird that the file is not there anymore. I hope algolia will remove the page from the search when we don't link it anymore.

PR #356, opened 2016-04-14
[2.2] Document relationship title format

SahAssar:
As implemented here:
https://github.com/bolt/bolt/blob/release/2.2/app/view/twig/editcontent/_relationships.twig#L17


bobdenotter:
Thanks! :+1:

PR #358, opened 2016-04-15
Update installation.md

KeizerDev:
You should make the `serve` file executable too.

### DIFF
--- a/source_docs/installation.md
+++ b/source_docs/installation.md
@@ -22,7 +22,7 @@ doesn't already exist. Enter the directory, and execute the following commands:
 ```bash
 curl -O http://bolt.cm/distribution/bolt-latest.tar.gz
 tar -xzf bolt-latest.tar.gz --strip-components=1
-chmod -R 777 files/ app/database/ app/cache/ app/config/ theme/ extensions/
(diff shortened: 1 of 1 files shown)


GawainLynch:
This is a developer option though, not really sure it belongs in end user documentation/installs enabled by default.

KeizerDev:
Why do you ship it in the `distribution/bolt-latest.tar.gz` then?

GawainLynch:
I wouldn't if it was up to me, I wouldn't even have it in the git repo :wink: 

Opinions @bobdenotter ?

bobdenotter:
I think there's very few people that use the zip or tgz distributions, that use the `serve` tool.

I could live with both options: chmod+x it, or remove it from the distro. Either would be fine with me, so let's see what others say.

GawainLynch:
@bobdenotter: My gut feeling is trim it from the repo and add it to the documentation… but yeah, if you want we can add it to Tuesday's agenda.

KeizerDev:
It actually runs as a service, I don't like that either

DennisSnijder:
I don't use the serve file either

GawainLynch:
Being removed in https://github.com/bolt/bolt/pull/5227

PR #365, opened 2016-04-20
Update install-command-line.md

evertalbers:
Hope the English is OK. 

I was pondering about adding something like 'go to the public directory and from go there one level up' which is clarifying to me but may cause confusion for people with a different background.


GawainLynch:
Manually merged.

PR #366, opened 2016-04-20
Update installation.md

evertalbers:
As obvious as it may seem, I completely overlooked the first option because of the 3 bullets below.

### DIFF
--- a/source_docs/installation/installation.md
+++ b/source_docs/installation/installation.md
@@ -4,13 +4,13 @@ Installing Bolt
 This page explains the various ways of installing Bolt. You can either use the
 command-line or your FTP-client to install it.
 
-There are several methods to install Bolt, but we recommend the "easiest way"
+There are several methods to install Bolt, but we recommend **the "easiest way"**
 [using the server command-line](install-command-line).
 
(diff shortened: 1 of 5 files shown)


source_docs/installation/installation.md
```diff
@@ -4,13 +4,13 @@ Installing Bolt
 This page explains the various ways of installing Bolt. You can either use the
 command-line or your FTP-client to install it.
 
-There are several methods to install Bolt, but we recommend the "easiest way"
+There are several methods to install Bolt, but we recommend **the "easiest way"**
 [using the server command-line](install-command-line).
```
GawainLynch: None of the three points below is called "easiest"… maybe we should either:
  * change "easiest" to "traditional"
  * change "traditional" to "easiest"

evertalbers: ...which means that you misread the page as well, it's **4** options altogether (CLI, FTP, Developers, Contributor).

Maybe we should then  put all options in the bullets or an ordered list to put them all in the same list.

GawainLynch:
Manually merged.

PR #374, opened 2016-04-21
Fix "quick links"

bobdenotter:


CarsonF:
I'm not understanding the change for `current` (slug) twig variable. I'm pretty sure that will break things

bobdenotter:
@CarsonF Before I changed it, and i did `{{ current }}` in the index.twig, It came out as `/local//local/foo/bar`.

CarsonF:
And now it comes out as `/local/foo/bar`?

bobdenotter:
Yes. At least on my setup it does. If you think it breaks, I'll do more monkey testing tomorrow. 🐵

CarsonF:
No maybe it changed with an update last night. I'll double check later tonight. I'll merge if we are good (which I think we are)

GawainLynch:
Let's do this!

PR #391, opened 2016-05-09
Using composer to create project

nnussbaum:
Using composer « create-project » to create project instead of « git
clone » so you get the latest stable version automatically


GawainLynch:
I can live with this :wink: :+1:

PR #400, opened 2016-05-13
Clarify slug generation for #362

### LINKED ISSUE #362: Clarification for singular/plural of slug and similar terms needed
Working with bolt a few years now and i am constantly tripping over this issue:

In bolt contenttypes, we have **name**, **singular_name**, **slug**, **singular_slug**, **table_name** and maybe even more. Some of these values are auto generated from others (sometimes changing from singular to plural or other way round), some methods support both quantities and some don't.
The documentation only has one short sentence in the contenttypes section for slug and singular_slug, telling us that they can be auto-generated and affect stuff.

Confusion with quantity here leads to problems with urls having wrong quantity, fetch functions returning empty results, labels confusing editors, broken relations, extensions not working correctly  and troubles with manual database changes (ex: in migration scripts).

**So the questions needing clarification are**:

1) Which of the 5 fields should be specified in what quantity (singular or plural)
2) which of these fields can be auto-generated and from what other field
3) Which functions (especially `setcontent`) support both quantities
4) Which field and in what quantity is used in what internal logic (especially in the database)
5) And a small side question: The key of a contenttype inside the contenttypes.yml doesnt seem to be used for anything right?

I would gladly help with writing additions to the docs here, sadly i still cant fully answer these questions.

SahAssar:
Fixes #362

### DIFF
--- a/docs/content/contenttypes-and-records.md
+++ b/docs/content/contenttypes-and-records.md
@@ -237,13 +237,14 @@ The available options are:
     the browser. It should be plural, if possible.
   - `singular_name`: The name of one Record in the Contenttype. This should be
     singular. For example, if the Contenttype's name is 'Pages', this should be
-    'Page'
+    'Page'.
   - `slug` (optional): This determines the slug of the contenttype, and
     therefore the URLs that are generated for this contenttype. When omitted,
-    the slug will be automatically generated.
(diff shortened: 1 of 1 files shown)


GawainLynch:
:+1:

PR #401, opened 2016-05-13
Fix references to error in case of non-existent template

SahAssar:
A non-existent template won't generate an error, it will just fall back to previous candidates.


bobdenotter:
Thanks! 👍

PR #402, opened 2016-05-13
update 2.2 docs

anketwokings:
- a lot of broken anchors
- download was for bolt-latest, which is the 3.0 branch at this time. Downloading 3.0 and trying to configure as a 2.2 is asking for trouble
- downloadlink was no https -> broken downloadlink

Idea: Also release a bolt-2-latest.zip, for easy of linking in these docs and downloading-links?


bobdenotter:
giphy 1

PR #422, opened 2016-06-09
Fix typo on composer.json assets definition

nbehier:


bobdenotter:
Good catch, thank you! 👍

PR #426, opened 2016-06-19
Fixed wrong call to findBy in findOneBy example

padawan:
The example code for findOneBy was using findBy instead of, err… findOneBy.


GawainLynch:
:+1:

PR #427, opened 2016-06-20
Update unit-testing.md

zomars:
Removed duplicate text, and removed stray tag.


SahAssar:
:shipit:

SahAssar:
These buttons are all so confusing!

PR #432, opened 2016-06-25
"Proxy .htaccess" doesn't rewrite /bolt-public/...

Sphinxxxx:
The `RewriteCond %{REQUEST_URI}` must include the leading slash on the `/public/` folder. If not, it will happily match urls to the standard `/bolt-public/` folder, and *not* do the needed rewrite.

The extra regex is apparently needed because Apache sometimes doesn't include the first slash in `REQUEST_URI`. More info here:
http://webmasters.stackexchange.com/questions/27118/when-is-the-leading-slash-needed-in-mod-rewrite-patterns


bobdenotter:
Good catch! Thanks!

PR #443, opened 2016-07-05
Added note regarding Twig and config.yml namespaces

Maelstrome26:
Added an explanation where Twig may need to use the `general` namespace of the config if the user added the variable directly into `config.yml`.


bobdenotter:
Thanks! 👍

PR #429, opened 2016-06-21
Add note about local assets to fix #4047

SahAssar:
Fixes bolt/bolt#4047

PR #444, opened 2016-07-10
Smaller *is* better, see #423 vs. #429

SahAssar:
Better to link to one place that explains it than having multiple places to update.

See #423 vs. #429


GawainLynch:
Love your work.

PR #445, opened 2016-07-10
fix botched link in last PR.

SahAssar:


GawainLynch:
"Won't somebody think of the commit count!"  :laughing: 

I totally think the :koala: snuck that last one in!

bobdenotter:
If you omit the `3.0/`, it'll stay working for future versions. :-)

bobdenotter:
ah, too late. :-)

PR #446, opened 2016-07-13
Added an example for randomising results with setcontent

Maelstrome26:


SahAssar:
closing cause it's documented here: https://docs.bolt.cm/3.0/templates/templatetags#setcontent

PR #451, opened 2016-07-24
Make sure people get it's web root, not root

SahAssar:


SahAssar:
Because I directed someone to this passage on slack and realized that it wasn't 3.0-ish enough...

bobdenotter:
👍

Issue #452, opened 2016-07-24
Document: Add the ability to skip over certain directives #5425


GawainLynch:
See https://github.com/bolt/bolt/pull/5425

Assigned to @Raistlfiren

Raistlfiren:
@GawainLynch This can be closed ^^ :)

bobdenotter:
👍

PR #453, opened 2016-07-26
Storage Additions

Raistlfiren:
Here is some documentation about the Bolt V3 layer. This might be something that @rossriley may want to go over. This is also in reference to https://github.com/bolt/docs/issues/452


GawainLynch:
Awesome job :+1:

PR #466, opened 2016-08-09
Remove inference to singular/multiple in setcontent from docs

SahAssar:
See bolt/bolt#5067


GawainLynch:
Question, is this a valid fix for v2.2 docs as well? If so I'll re-target the branch

SahAssar:
Yeah. you're right, this should go into 2.2

GawainLynch:
Merged into 2.2 and cascaded via merge … and deployed :+1:

PR #473, opened 2016-08-24
Tweak "note" to indicate use of theme.yml

GawainLynch:


SahAssar:
Um, shouldn't that be contenttypes.yml, not theme.yml?

GawainLynch:
Maybe… that's part of the freaking problem :laughing: this stuff is in need of love as things do not connect well.

Let's talk about it when you get here, no rush on this one

PR #489, opened 2016-09-10
change composer create-project to composer install command

PhillippOhlandt:
This makes much more sense than before.

### DIFF
--- a/docs/installation/composer/creating-a-new-project.md
+++ b/docs/installation/composer/creating-a-new-project.md
@@ -163,7 +163,7 @@ Finally run Composer to install the required libraries and configure the
 installation:
 
 ```
-composer create-project
+composer install
 ```
 
 ### Permissions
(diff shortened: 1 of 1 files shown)


GawainLynch:
That isn't correct. We need the events to run.

PhillippOhlandt:
But `composer create-project` isn't correct too. 

Then we should move 
```
"Bolt\\Composer\\ScriptHandler::installThemesAndFiles",
"nut extensions:setup"
```

to `post-install-cmd` and we could remove `Bolt\\Composer\\ScriptHandler::configureProject` completely.

Then the required still runs after you ran `composer install`.

PhillippOhlandt:
Note: I mean just the event handlers in the `composer.json` in the docs here.

PhillippOhlandt:
Other thing. I would prefer to remove the whole "Create all config files manually and run composer *" section anyway. 

`composer create-project bolt/composer-install` is the preferred way and the manual method means just trouble.

GawainLynch:
:+`

PR #490, opened 2016-09-11
Updated Custom Bootstrapping Guide

rossriley:
updated bootstrapping guide to reflect new modifications in project structure

### DIFF
--- a/docs/extensions/custom-bootstrapping.md
+++ b/docs/extensions/custom-bootstrapping.md
@@ -5,7 +5,7 @@ Creating your own Bootstrap
 =============================
 
 Whilst Bolt is designed to be simple for anyone to install, its core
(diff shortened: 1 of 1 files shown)


PhillippOhlandt:
Why doesn't this point to `release/3.0`?

rossriley:
Most of it is specific to https://github.com/bolt/bolt/pull/5748 which will target 3.2

PhillippOhlandt:
And whats with 3.0 and 3.1? I think we need something for that too.

GawainLynch:
@rossriley the difference being the hidden fail versus non-hidden, isn't it? In which case we can probably merge to 3.0 an updated version of this PR, and then on  cascade update the 3.2+ branch(es)

GawainLynch:
Covered in #530

Issue #493, opened 2016-09-15
Add some words about translations inclusion on extension


nbehier:
If I well understood, in Bolt 3.1, translations will be automatically included in extension. Could you add some word on the doc about it
https://github.com/bolt/bolt/blob/release/3.1/src/Extension/TranslationTrait.php

GawainLynch:
@Raistlfiren I don't suppose you have some spare time to follow up with docs for that PR? If not, let us know, as @nbehier is right — it'd be good to have at least the basics — and the task should be allocated otherwise :smile:

nbehier:
@Raistlfiren @GawainLynch @PhillippOhlandt You're awesome guys! Thanks!

PR #495, opened 2016-09-15
Added Translations documentation.

Raistlfiren:
This is a real basic overview of how to use Translations with Extensions. Simple but sweet... 

Closes https://github.com/bolt/docs/issues/493

### DIFF
--- a/docs/extensions/basics/index.md
+++ b/docs/extensions/basics/index.md
@@ -9,4 +9,5 @@ pages:
     - twig
     - web-assets
     - testing
+    - translations
 ---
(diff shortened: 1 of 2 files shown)


PhillippOhlandt:
Looks good structure wise but I don't think we need to mention the Bolt version because it's only in the 3.1+ docs.

Raistlfiren:
Well, the translation functionality wasn't included until 3.1 so I would hate for someone to overlook that minor fact.

PhillippOhlandt:
Yeah but for that reason, we have versioned documentation.

Raistlfiren:
I just feel like there will be a lot of headaches wondering "why is this not working??". :) OH well.

PhillippOhlandt:
Yeah but that's not our problem. We also have some additional pages for the storage system on the 3.1 or 3.2 branch.

GawainLynch:
:+1:

PR #496, opened 2016-09-18
Field type fixes

PhillippOhlandt:
First I replaced some code lines with code blocks for consistency. 

Then I also changed all the field names in the example with something more useful and real world'ish. 

I also fixed some examples, where the field name in the twig example was not the same as the field name in the yaml definition.

Sorry for the commit spam, I just wanted to make it as easy as possible for git for the merge down to 3.0+.


PhillippOhlandt:
Every editor handles that differently and it looks weird and wrong in the docs. We also don't indent all code examples.

PhillippOhlandt:
And not all were intended so it looked like a copy&paste issue.

PhillippOhlandt:
Every editor handles that differently and it looks weird and wrong in the docs. We also don't indent all code examples.

GawainLynch:
@bobdenotter & @SahAssar … any opinions here?

SahAssar:
I'd prefer indented for the copy-paste ease...

The rest of it I'm all for though, I renamed all the examples to `name` when I moved them just because some people might have previously assumed that every field needed to be named after the fieldtype since they were named that way in the old examples.

bobdenotter:
I agree with @SahAssar: It should be "copy-pasteable" :-)

SahAssar:
👍

bobdenotter:
@GawainLynch Your "requested changes" are resolved now, right?

GawainLynch:
WFM2 :+1:

PR #503, opened 2016-09-23
Change 'contenttype' to 'ContentType' in texts

anketwokings:
please check, I think I did all that needed to be done, but second pair of eyes is appreciated.


GawainLynch:
Merged into `release/3.0` via 01d1bf0ab44a8281abdd18dae64d988f339d66b1

PR #510, opened 2016-09-30
Add google api key info to cheatsheet geolocation field

anketwokings:
because i can.
And yes, probably not the right version. Already saying sorry to Phillipp and making offerings of beer to the version-gods.


GawainLynch:
Merged into `release/3.0` vai 3cf2c9a827d0619954ac92a430c4edae97900804

Thanks @anketwokings :+1:

PR #513, opened 2016-10-06
remove moved directories from default config

jadwigo:
the mentioned directories are moved outside the webroot - so they only interfere with legitimate site content.

### DIFF
--- a/docs/installation/webserver/nginx.md
+++ b/docs/installation/webserver/nginx.md
@@ -71,11 +71,6 @@ server {
         deny all;
     }
 
-    # Block access to the app, cache & vendor directories
-    location ~ /(?:app|src|tests|vendor)/(.*)$ {
-        deny all;
-    }
-
     # Block access to certain JSON files
     location ~ /(?:bower|composer|jsdoc|package)\.json$ {
         deny all;


GawainLynch:
Not really sure completely removing it is the right approach … but close to. Ambivalent about this one though.

Technically they are out of the web root in nearly all cases, the general exception being people stuck on crappy hosting … and that's all Apache.

@CarsonF & @bobdenotter

CarsonF:
My uninformed view is anyone using nginx knows what they are doing..therefore doesn't have their web root the same as their project root.

jadwigo:
Also - a few lines earlier executing php files in the apps, extensions and other directories is excluded anyway by
```
    # Block PHP files from being run in upload (files), app, theme and extension directories
    location ~* /(?:app|extensions|files|theme)/(.*)\.php$ {
        deny all;
    }
```

SahAssar:
I think it should stay. Making an assumption that just because someone is use certain software they will be competent is dangerous and if someone uses a git install (I know that is discouraged for non-dev work, but still) the folders will be there.

CarsonF:
Here's a novel idea: We could change our git repo to not have the web root be the same as the project root. smh

SahAssar:
Sure, but until that change is in perhaps we should keep this in?

jadwigo:
If we keep these lines we also need to tell users that any slug may not contain app/ src/ tests/ vendor/ ... because those will give you a HTTP 403 in nginx

In a normal install these lines will actually decrease the usability of bolt. It will for example prevent content types with the slug apps .. because the detail page http://example.com/app/pokemon-go will be forbidden and it will also block any javascript in a vendor or src folder like some frameworks like to do.


```
    # Block access to the app, cache & vendor directories		
    location ~ /(?:app|src|tests|vendor)/(.*)$ {		
        deny all;		
    }
```

And because it is a webserver 403 it is nigh impossible to debug.

The actual directory structure has changed, removing the corresponding lines places the burden on the people who move their webroot to something not standard instead of the users who don't move stuff around.

Keeping these lines in will make it harder for people who install bolt the recommended way on nginx. 

Also this is just copy pasted from the bolt 2 config - and keeping copypasta around because it's already there is not good.😕

GawainLynch:
> The actual directory structure has changed, removing the corresponding lines places the burden on the people who move their webroot to something not standard instead of the users who don't move stuff around.

Yes, and no. Un-commented, I'll give you that as a point. But the plan all along has been to update these so it is a bit more informative … especially when it comes to securing your site. 

> Also this is just copy pasted from the bolt 2 config - and keeping copypasta around because it's already there is not good.

Yes … and I am to blame, as I have had a TODO forever … but my hand was forced on a release date and I haven't stopped since … but if I was to take the suggested approach with core code, I'm pretty sure I'd lose the respect of our fellow team members — and commit access — pretty quickly. :wink:

GawainLynch:
OK, how about this for a compromise?
```
    # Block access to the app, cache & vendor directories
    #
    # NOTE: If you do *not* have your web root as a sub-directory of your 
    # installation root, you should uncomment this location block to prevent
    # site visitors having access to the various directories that contain
    # executable code.
    #
    # NOTE: This approach is *not* recommended for production use.
    #
    #location ~ /(?:app|src|tests|vendor)/(.*)$ {
    #    deny all;
    #}
```

SahAssar:
WFM 👍

jadwigo:
I would rephrase that a little ... 
```
    # Block access to the app, cache & vendor directories
    #
    # NOTE: If you have one or more of 'app' , 'src', 'test' and 'vendor' as 
    # sub-directories of your installation root, you should uncomment
    # this location block to prevent site visitors having access to the 
    # various directories that contain executable code.
    #
    # NOTE: This approach is *not* recommended for production use.
    #
    #location ~ /(?:app|src|tests|vendor)/(.*)$ {
    #    deny all;
    #}
```

GawainLynch:
> I would rephrase that a little ...

I like *much* more than mine! 

Are you good to update it, or would you like me to ninja it in on the branch rebase?

jadwigo:
ninja away :)

GawainLynch:
Merged into `release/3.0` via 3201d649b84a021aad30177950bc41f907562413

PR #529, opened 2016-10-15
Fix options in file, filelist, image and imagelist field types

### LINKED ISSUE #528: Incorrect option in Image Field Docs
As seen here: https://docs.bolt.cm/3.1/fields/image

As far as I can see we have never supported `accept_file_types` as an option at the field level, this should instead be `extensions` the former is correct in the general config, but the latter at field level.

Edit: looks as though this applies to all file field docs: `file`, `filelist`, `image` and `imagelist`

PhillippOhlandt:
fixes #528

### DIFF
--- a/docs/fields/file.md
+++ b/docs/fields/file.md
@@ -24,5 +24,11 @@ Simple file upload/select field.
 The field has a few options to change the appearance and functionality of the
 field.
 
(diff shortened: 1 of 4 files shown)


GawainLynch:
:+1:

PR #530, opened 2016-10-16
Update bootstrapping guide

GawainLynch:
This PR is aimed at superseding #490, and also finally documents the `application` parameter that can be used in the loader configuration file (`bolt.yml` or `bolt.php`).

We are trying to also point to a more modular approach to our loading/configuration that has been worked on heavily by the team collectively, and obviously there is still work to be done in core.

Something that this effort by @CarsonF and myself is about, is trying to make some initial steps in improving the communication to all, by those actually doing the initial core work.

What I hang my head in shame about right now, is that this supersedes @rossriley's work on #490, and he's probably the best of all of the back-end team members at documenting stuff, and he's put in a fair amount of effort, but only missed because the two people behind *this PR* didn't initially communicate well/at all their changes to core.

### DIFF
--- a/docs/extensions/custom-bootstrapping.md
+++ b/docs/extensions/custom-bootstrapping.md
@@ -1,89 +1,169 @@
 ---
 title: Custom Bootstrapping
 ---
-Creating your own Bootstrap
-=============================
+Customising Bootstrapping
+=========================
 
-Whilst Bolt is designed to be simple for anyone to install, its core
-functionality is also modular and easy to configure for anyone who is
-comfortable making a few modifications to their bootstrap code.
+Whilst Bolt is designed to be simple for anyone to install, its core 
+functionality is also modular and easy to configure for developers who are
+comfortable making a few modifications to how Bolt is bootstrapped.
 
-By default Bolt ships with its own bootstrap file which is responsible for
-setting up and running a Bolt application. Most of the advanced configuration
-options are best attained by taking over this responsibility in your own
-application. In the root of your project you will see an `index.php` file that
-should look like this:
+There are two ways your Bolt application is used, via web requests and via the
+command line with `nut`. Because there are two entry points, bootstrapping is
+done in a common `bootstrap.php` file. 
 
-```
-$app = require dirname(__FILE__) . '/../vendor/bolt/bolt/app/web.php';
-if ($app === false) {
-    return false;
-}
-$app->run();
-```
+One of the important functions of the bootstrap file is to read a loader 
+configuration file in the root directory of your project. By customising this
+configuration file, you can modify things like paths, the application class to
+use, and other services used by the application. This file can be either 
+`.bolt.yml` or `.bolt.php`. 
 
-Ideally you would replace the path to bootstrap.php with a link to your own
-bootstrap file, by convention it is best to name this `.bolt.php` since that filename
-is looked for automatically by other Bolt components. For example:
-
-```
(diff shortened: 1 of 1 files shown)


GawainLynch:
Also noting, that I will take care of the hidden file changes in 3.2+, if/when this lands, during the branch merges.

docs/extensions/custom-bootstrapping.md
```diff
@@ -1,89 +1,169 @@
...
+done in a common `bootstrap.php` file. 
 
-```
-$app = require dirname(__FILE__) . '/../vendor/bolt/bolt/app/web.php';
-if ($app === false) {
-    return false;
-}
-$app->run();
+One of the important functions of the bootstrap file is to read a loading 
+configuration file in the root directory of your project. This configuration is
```
PhillippOhlandt: ` This configuration is you can leverage customisation of things like paths, the application where class, and services used by the application.`

Seems like words are missing in this sentence? Also, `the application where class` doesn't sound correct.

docs/extensions/custom-bootstrapping.md
```diff
@@ -1,89 +1,169 @@
...
+example.
 
-$app->initialize();
-$config = [
-    'application' => $app,
-    'resources' => null,
+```php
+<?php
+
+// However you need create the application.
```
PhillippOhlandt: Is that correct English? Feels weird. Perhaps it should be `However you need to create the application.`?

PhillippOhlandt:
Also, could we have a guide about setting up PHPUnit within a custom bootstrap? Writing tests on extensions is easy because PHPUnit is already properly configured there if you use the example repo or Günther. But here I am not sure how to set it up.

But that can be done in a later PR and is not required now. Just a wish from my side :)

GawainLynch:
> Also, could we have a guide about setting up PHPUnit within a custom bootstrap

As part of this no. Out of scope, and our approach to date has been around functional not unit tests

docs/extensions/custom-bootstrapping.md
GawainLynch: You are correct! Spotting a missing infinitive verb … your English is getting pretty damn good mate! :+1:

docs/extensions/custom-bootstrapping.md
GawainLynch: Yeah, I blame … um … well, not the :koala: & not the :hamster: … who can we needlessly label with blame here? :wink: 

I'll fix soon

PhillippOhlandt:
> As part of this no. Out of scope

This was only an additional wish from my side :)

>  and our approach to date has been around functional not unit tests

Yeah, but when I build my application, I would like to write unit tests :D

docs/extensions/custom-bootstrapping.md
PhillippOhlandt: Nah, just me not understanding the shit and trying to give the sentence at least a bit of sense :P

GawainLynch:
@PhillippOhlandt Updated mate, thanks so much for the review!

PhillippOhlandt:
Are we GTG?

GawainLynch:
There has been zero comment, or decent … So …

PhillippOhlandt:
I initially thought @SahAssar or @bobdenotter want to check too. But if they don't complain I will merge it in.

PhillippOhlandt:
wush

PR #535, opened 2016-10-19
Site Updates

### LINKED ISSUE #506: Request: Have listings in 'overview' pages.
It'd be very helpful to users if pages that have 'children' would show those pages in the right hand side. Not just in the navigation tree. 

For example on pages like this: 

https://docs.bolt.cm/3.1/internals

It should look like this: 

screen_shot_2016-09-23_at_16_33_57

CarsonF:
See commits for info.

Fixes #506

Warning: Cache needs to be cleared after pulling this change.


GawainLynch:
No :koala: spotted … let's see how this goes …

PR #545, opened 2016-10-26
Add Sitemap.xml functionality.

### LINKED ISSUE #544: Sitemap.xml. The documentation site should have it.
For better indexing by the googles and such.

bobdenotter:
Fixes #544

### DIFF
--- a/src/Controllers.php
+++ b/src/Controllers.php
@@ -122,7 +122,7 @@ class Controllers implements ControllerProviderInterface
     public function sitemap(Version $version)
     {
         $twigVars = [
-            'sitemap' => $version->getPage(''),
+            'root' => $version->getPage(''),
         ];
 
         $xml = $this->render('sitemap.twig', $twigVars);
--- a/src/Controllers.php
+++ b/src/Controllers.php
@@ -33,6 +33,13 @@ class Controllers implements ControllerProviderInterface
             ->value('page', '')
             ->assert('page', '.*');
 
+        $ctr->get('/sitemap/{version}.xml', [$this, 'sitemap'])
+            ->bind('sitemap')
+            ->value('version', '');
+
+        $ctr->get('/sitemap.xml', [$this, 'sitemap_list'])
+            ->bind('sitemap_list');
+
         $ctr->convert('version', function ($version) {
             if (!$version) {
                 return $version = $this->app['documentation']->getDefault();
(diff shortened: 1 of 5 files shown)


src/Controllers.php
```diff
@@ -106,6 +113,40 @@ public function cheatsheet(Version $version)
...
+     * Controller for sitemap (single)
+     *
+     * @param Version $version
+     *
+     * @return mixed
+     */
+    public function sitemap(Version $version)
+    {
+        $twigVars = [
+            'menu' => $version->getMenu(),
```
CarsonF: Could you use `$version->getPage('')` here instead. I'm actually trying to kill `getMenu`.

Maybe just change the key to `page` or `root` or `rootPage`.

view/templates/_macros.twig
```diff
@@ -56,3 +56,11 @@
         </ul>
     </li>
 {% endmacro %}
+
+{% macro sitemap(page) %}
+    {% from _self import sitemap %}
+    <url><loc>https://docs.bolt.cm{{ page.url }}</loc></url>
+    {% for sub in page.children if sub is not empty %}
```
CarsonF: The `if` isn't needed here. And would change to `page.subPages` with menu change.

src/Controllers.php
```diff
@@ -106,6 +113,40 @@ public function cheatsheet(Version $version)
...
+
+    /**
+     * Controller for sitemap-sitemap (multiple)
+     *
+     * @return mixed
+     */
+    public function sitemap_list()
+    {
+        $twigVars = [
+            'versions' => array_keys($this->app['documentation']->getVersions()),
```
CarsonF: `array_keys` isn't needed here. `Version` has a `__toString` method.

src/Controllers.php
```diff
@@ -33,6 +33,13 @@ public function connect(Silex\Application $app)
             ->value('page', '')
             ->assert('page', '.*');
 
+        $ctr->get('/sitemap/{version}.xml', [$this, 'sitemap'])
+            ->bind('sitemap')
+            ->value('version', '');
```
CarsonF: `/sitemap/.xml`? I don't think you can or need to set version's default value to empty string here.

view/templates/_macros.twig
```diff
@@ -56,3 +56,11 @@
         </ul>
     </li>
 {% endmacro %}
+
+{% macro sitemap(page) %}
+    {% from _self import sitemap %}
+    <url><loc>https://docs.bolt.cm{{ page.url }}</loc></url>
```
CarsonF: Do `loc`'s need to be absolute?

src/Controllers.php
bobdenotter: Can I remove this one too, then?

https://github.com/bolt/docs/blob/site/src/Controllers.php#L32

CarsonF: No because that indirectly defines the homepage.

bobdenotter:
@CarsonF All your suggestions should be addressed now. :-)

bobdenotter:
Well, It _is_ the tree with links to all of the pages, so I figured the name was apt.. 

I'll change it though, 10 second fix. :-)

PR #548, opened 2016-10-29
[RFC] Improving the 'version select dropdown'.

bobdenotter:
Before: 

screen shot 2016-10-29 at 17 23 09

After: 

screen shot 2016-10-29 at 17 20 54

 - The label makes more sense
 - You can see what is the current stable release, and which is "future release"
 - If the `var/versions/` folder contains non-numeric folders, list them at the bottom.


bobdenotter:
Small change, reversed the order: 
screen shot 2016-10-29 at 17 29 10

PR #552, opened 2016-10-31
Add thumbnails configuration page

PhillippOhlandt:
Preparation for adding documentation about thumbnail aliases in 3.2.

Feel free to suggest text changes :)

### DIFF
--- a/docs/configuration/thumbnails.md
+++ b/docs/configuration/thumbnails.md
@@ -30,14 +30,14 @@ Settings
 
 | Setting | Explanation |
 | --- | --- |
-| default_thumbnail | The default size of images, when using `{{ record.image|thumbnail() }}`. |
-| default_image | The default size of images, when using `{{ record.image|image() }}`. |
-| quality | Should be between `0` (horrible, small file) and `100` (best, huge file). |
-| cropping | One of either `crop`, `fit`, `borders` or `resize`. |
-| notfound_image | Path to the image that will be shown if the thumbnailer can't find the given image. |
-| error_image | Path to the image that will be shown if an error occurs during the thumbnail generation. |
-| save_files | Can be either `true` or `false`. If true, it saves generated thumbnails to disk so they don't need to be generated again. |
-| allow_upscale | Can be either `true` or `false`. If true, the thumbnailer can make images bigger than the original. |
-| exif_orientation | Can be either `true` or `false`. If true, the thumbnailer will use the image orientation from its EXIF data. |
-| browser_cache_time | Sets the amount of seconds that the browser will cache images for. Set it to activate browser caching. |
+| `default_thumbnail` | The default size of images, when using `{{ record.image|thumbnail() }}`. |
+| `default_image` | The default size of images, when using `{{ record.image|image() }}`. |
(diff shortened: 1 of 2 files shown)


docs/configuration/thumbnails.md
```diff
@@ -0,0 +1,43 @@
...
+    default_thumbnail: [ 160, 120 ]
+    default_image: [ 1000, 750 ]
+    quality: 80
+    cropping: crop
+    notfound_image: view/img/default_notfound.png
+    error_image: view/img/default_error.png
+    save_files: false
+    allow_upscale: false
+    exif_orientation: true
+#    browser_cache_time: 2592000
```
GawainLynch: Question to the audience: Do we still use `browser_cache_time`

PhillippOhlandt: I took it from the configuration :D

PhillippOhlandt:
@bobdenotter your turn :)

docs/configuration/thumbnails.md
bobdenotter: I don't think it's used anymore: 

```
$ ag 'browser_cache_time' src                                 
src/Provider/ThumbnailsServiceProvider.php
69:        $app['thumbnails.cache_time'] = $app['config']->get('general/thumbnails/browser_cache_time');

$ ag 'thumbnails.cache_time' vendor/bolt/thumbs         
vendor/bolt/thumbs/src/ServiceProvider.php
31:                $app['thumbnails.cache_time']
53:        $app['thumbnails.cache_time'] = null;
```

Should we just remove it? (both here, and in the source code)

bobdenotter:
I really like he tables. We should definitely use more of them! :-)

Only thing is, they need to be linearized, because currently they break the layout on mobile: 

screen shot 2016-11-01 at 07 06 44

Perhaps "stack" them on smaller screens? http://foundation.zurb.com/sites/docs/table.html#stacked-table

docs/configuration/thumbnails.md
bobdenotter: If it's only _set_ and then nothing is done with it, does that count as "being in use"? Or am I missing something? 

(not saying we _should_ remove it, just curious)

GawainLynch: The service provider [passes that value](https://github.com/bolt/bolt-thumbs/blob/master/src/ServiceProvider.php#L31) into the [Responder](https://github.com/bolt/bolt-thumbs/blob/master/src/Responder.php#L56) class and then uses it as a parameter to the [caching call](https://github.com/bolt/bolt-thumbs/blob/master/src/Responder.php#L97).

bobdenotter: Ah, duh!

GawainLynch: Yeah, I has a similar "duh" moment around my original question … was just in too much of a hurry yesterday to do 2 seconds worth of work, that cost you and I far more than that in the end … 

Lesson to be learned there, Gawain? :wink:

GawainLynch:
FWIW :+1: from me

PhillippOhlandt:
Ok, changing the parameters to be `code`. I had that before, It just looked less clean.

bobdenotter:
Looks good to me! 

🎏

PR #571, opened 2016-11-09
[3.2] Added example usage of Thumbnail alias

Maelstromeous:


PhillippOhlandt:
It is already documented here https://docs.bolt.cm/3.2/templating/templatetags#thumbnail

PR #573, opened 2016-11-10
HTTP Cache docs

GawainLynch:
Related to https://github.com/bolt/bolt/pull/6023

### DIFF
--- a/docs/performance/http-cache.md
+++ b/docs/performance/http-cache.md
@@ -1,0 +1,64 @@
+---
+title: HTTP Cache
+---
+HTTP Cache
+==========
+
+Bolt comes with [Symfony HTTP Cache][] that will act as a simple reverse proxy
+to cache rendered pages on disk for a period of time.
+
+This can provide a performance boost on some pages, as means less work for the
+werb server to do in order to serve the requested page repeatedly in a short
+amount of time.
+
+<p class="note"><strong>Note:</strong> Requests for logged on users will not
+be cached.
+</p>
+
+
+Enabling
+--------
+
+To enable [Symfony HTTP Cache][] for your Bolt install, in your `config.yml`
+file, find the `caching:` key, and enableing the `request` parameter.
+
+
+```
+caching:
+    request: true
+```
+
+
+Fine Tuning Options
+--------------------
+
+There are several options that get passed to the cache, this can be set under
+the `performance:` key in your `config.yml` file.
+
+```
+performance:
+    http_cache:
+        options:
+            default_ttl: 8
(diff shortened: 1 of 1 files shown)


GawainLynch:
I completely disagree.

This is going to be part of a larger set of documents, aimed at the expert group of users.

PhillippOhlandt:
But it's about a setting in config.yml and therefore needs to go into the config section. We wont spread stuff all over the docs again.

GawainLynch:
> But it's about a setting in config.yml and therefore needs to go into the config section. We wont spread stuff all over the docs again.

And the other 99% that has to do with everything else? …and the Varnish follow up, that doesn't even touch Bolt?

docs/performance/http-cache.md
```diff
@@ -0,0 +1,64 @@
...
+title: HTTP Cache
+---
+HTTP Cache
+==========
+
+Bolt comes with [Symfony HTTP Cache][] that will act as a simple reverse proxy
+to cache rendered pages on disk for a period of time.
+
+This can provide a performance boost on some pages, as means less work for the
+werb server to do in order to serve the requested page repeatedly in a short
```
bobdenotter: They took our jerbs!

docs/performance/http-cache.md
```diff
@@ -0,0 +1,64 @@
...
+amount of time.
+
+<p class="note"><strong>Note:</strong> Requests for logged on users will not
+be cached.
+</p>
```
bobdenotter: Is this "logged on" users, or any user that has a session? Eg. from Members? 

Does it also automatically exclude POST requests for forms?

PhillippOhlandt:
Well, those Settings should still be on a site in the config section. You can link to it then. Let's talk about that when I am home. I think about a solution in the meantime

docs/performance/http-cache.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+Enabling
+--------
+
+To enable [Symfony HTTP Cache][] for your Bolt install, in your `config.yml`
+file, find the `caching:` key, and enableing the `request` parameter.
```
bobdenotter: enabling

docs/performance/http-cache.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+```
+caching:
+    request: true
+```
+
+
+Fine Tuning Options
```
bobdenotter: I think this should be [Fine-tuning](https://en.wikipedia.org/wiki/Fine-tuning), right?

PR #576, opened 2016-11-11
Setting the default to correct version 3.2.

bobdenotter:
[ Insert beaker.gif ]

### DIFF
--- a/src/Application.php
+++ b/src/Application.php
@@ -48,7 +48,7 @@ class Application extends Silex\Application
             $config = Yaml\Yaml::parse(file_get_contents(__DIR__ . '/../app/config.yml'));
         }
         $this['debug'] = isset($config['debug']) ? $config['debug'] : true;
-        $this['documentation.versions.default'] = isset($config['default-version']) ? $config['default-version'] : '3.1';
+        $this['documentation.versions.default'] = isset($config['default-version']) ? $config['default-version'] : '3.2';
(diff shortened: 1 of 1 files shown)


GawainLynch:
@CarsonF any reason this is hardcoded?

bobdenotter:
It picks it up from `app/config.yml`, but there should be _some_ default, right?

GawainLynch:
Yeah, just worries me that we'd be updating this for every minor/major version change

GawainLynch:
I'll send in, and we can come up with better ideas later … 

Just a note though … as we got our hand forced on the default branch, and it is currently `master`, people won't have `release/3.2` available locally without building worktrees … add to that, `release/3.2` is almost always going to be the wrong branch to work on … CBF Factor 11 :hushed: :see_no_evil:

bobdenotter:
The README [here](https://github.com/bolt/docs) details the set up. Even I could figure it out from that. I think it's not too much to expect from people who want to work on the docs, to read that.

GawainLynch:
> I think it's not too much to expect from people who want to work on the docs, to read that.

Not too much at all … just history has proven otherwise. Several times :wink:

CarsonF:
We could change the default to use the first version found? Or maybe the highest version found?

GawainLynch:
Meh, too much work :laughing:

PR #582, opened 2016-11-14
Make it work and change `/apigen/` links to `docs/api` links.

bobdenotter:


GawainLynch:
Hmm... Didn't know the apigen docs had made it here and have (from tired memory) been against it, as this is both not final URLs , and not wanting to declare theses as public just yet.

That said, before we do this we should define *what* URLs we want to use, as these ones aren't ideal, aren't rooted, and aren't (hopefully) indexed... Just worried about "why you 404, you suck" responses :laugh:

bobdenotter:
My next PR will remove them from this page: https://docs.bolt.cm/3.0/getting-started/introduction

... onto this page: https://docs.bolt.cm/3.0/core-development

If that makes it better. At least, they _for sure_ don't belong on the site-opener. :-)

We should probably _delete_ the old stuff at `/apigen` also :-)

GawainLynch:
WFM... On the iPad now, so shall deploy and delete old cruft in the morning ... I'll try and think about routes in the mean time too. Right now they're just thrown together for bookmarking by the :hamster: right now, but if we're going public we should have something more functional.

GawainLynch:
Should this be more of a "warning" colour?"
image

GawainLynch:
Drop down should show more, and this subtitles?
image

GawainLynch:
image

GawainLynch:
There is no closure called `BoltStorageFieldTypeTemplateFieldsType` and I can't fine teh `	/Users/bob/.composer/vendor/composer/ClassLoader.php` directory on my computer :stuck_out_tongue_winking_eye: 
image

GawainLynch:
Version switcher on teh generated docs is broken
image

GawainLynch:
Also, dev.bolt.cm is a "trash box" … and one more thing to support if we're using it publicly, which currently means more work for me as nobody else seems to have time.

bobdenotter:
> Should this be more of a "warning" colour?"

Will be fixed in a subsequent PR, when I revamp the "Info", "Note", "Warning" and "Advanced" callouts.

> Drop down should show more, and these "Show" & "entries" subtitles?

Fixed. That was the pager. I turned it off, because we show them all. 

> There is no closure called BoltStorageFieldTypeTemplateFieldsType and I can't find the /Users/bob/.composer/vendor/composer/ClassLoader.php directory on my computer 😜 

I double checked, and I can confirm it's there. Did you look in the correct folder? 

Kidding aside, I'll regenerate those tables, and PR updates to the relevant `3.x` branches. 

> Version switcher on the generated docs is broken

That should be fixed on `/api/docs`, right? 

FWIW, I think it's valuable, and would fit in nicely with our plans to provide "advanced" docs to advanced users. I can put in my share of effort on the docs.bolt.cm side of things, too. :-)

PR #601, opened 2016-12-07
Remove reference to 'is defined' for extensions

GawainLynch:


bobdenotter:
Is there another mechanism we can use to check if a certain extension exists, these days?

PR #602, opened 2016-12-07
Updating class_reference for 3.3

bobdenotter:
Only relevant for 3.3 and up.


bobdenotter:
@GawainLynch No rush with this one.. Going forward, should I alphabetize them to minimize delta's?

GawainLynch:
Yeah, if sorting can be done programatically when you generate those, sure! Makes everything easier.

PR #606, opened 2016-12-08
Replacement for "defined for extensions", for 3.3+ only.

bobdenotter:
Only for 3.3 and up, because of the nasty `template_from_string` hack. Thanks, @carsonf 😘

### DIFF
--- a/docs/templating/templatetags.md
+++ b/docs/templating/templatetags.md
@@ -785,6 +785,39 @@ Examples:
 {% endif %}
 ```
 
+### defined (for extensions)
+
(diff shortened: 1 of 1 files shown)


bobdenotter:
See also: https://github.com/bolt/bolt/issues/5372

bobdenotter:
@carson said this would be hard to do it in a cleaner way, so I think we should go with my hackish suggestion for now.

GawainLynch:
WFM then

bobdenotter:
One of those things like "One day, when we have nothing to do, let's philosophise about a neat and clean way to do this" 

🤣

PR #610, opened 2016-12-15
Fix extending global twig environment

maetthu:
When extending the global twig environment, it should be wrapped around $app->share() so it won't get called each time the service is consumed.


bobdenotter:
@maetthu Thanks! 

@GawainLynch Can you handle this one?

GawainLynch:
Somebody had their mind on Pimple 3 when they did this :laughing: 

Merged into `release/3.0` via 74bdf3cd3acee13c4a58c3d463a1dc865caa399d … thanks @maetthu

PR #623, opened 2017-01-06
Set correct names for composer .dist files

anketwokings:
To avoid any confusion, since after the installation you only have the .dist files


GawainLynch:
:+1:

PR #631, opened 2017-01-22
Change to more semantic link.

arkhi:
- Add the abbreviation for WYSIWYG.
- Apply a tip from the W3C: https://www.w3.org/QA/Tips/noClickHere.

PR #633, opened 2017-01-23
Some fixes and cleanup

arkhi:
I read [your previous comment](https://github.com/bolt/docs/pull/631#issuecomment-274316752) about versions 3.0 and 3.2, but I started some commits in 3.2 already, I don’t have much time these days, and I believe you’ll be faster to rebase properly based on your knowledge of the workflow. Sorry to add some work.

- `example below` and `following example`. This is about not assuming where some content will visually be, but where it will sit in the flow of the text: If the documentation is printed, for example, then some content at the top of next page is not below, but always following or preceding.
- I simplified a few sentences; sometimes removing text that I found redundant. I prefer simple sentences, so let me know if that was too much. :)
- Use identical naming for `drop-down list`.
- Add “missing” closing tag in HTML example.
- Add missing URL to Geolocation field documentation.


GawainLynch:
> I believe you’ll be faster to rebase properly based on your knowledge of the workflow. Sorry to add some work.

Zero stress mate, I'll get onto this later today :+1:

PR #640, opened 2017-01-30
Fix key variable being overwritten in loop

SahAssar:
When we set the key in the loop the we couldn't get the current saved value with context.content.get(key). Also use `|humanize` on the label instead of `|capitalize` on the saved key (wut?).


bobdenotter:
I didn't even know about `humanize` being a twig filter.. neat. :-)

PR #644, opened 2017-02-01
Updating the docs for Local Extensions in Bolt 3.3.

bobdenotter:


docs/howto/installing-local-extensions.md
```diff
@@ -5,34 +5,96 @@ level: intermediate
...
+For example, if the local folder does not exist, then create it and inside
+`extensions/local` make an additional folder, `[Extensionname]` and inside that folder add a valid Bolt extension class and save as `[Extensionname]Extension.php`.
+
+Your extensions folder would look something like this:
+
+```
+extensions
+├── local/
+│   └── Listing/
+│   │   ├── ListingExtension.php
```
GawainLynch: I am still wondering if we shouldn't keep the documented approach in line with how we do Composer based extensions, as I know for fact we get people installing things like BoltForms, or forks of certain extensions as locals. That way the documented migration path is nothing more than the autoloader settings in `composer.json`

bobdenotter: @GawainLynch Basically, "adding a `src/` folder in there", right? If so, i'll gladly add that.

GawainLynch: Well, it affects the structural layout, so a bit broader.

I would suggest:
  * we stick with Composer standard `namespace AuthorName\PackageName;` as a base
    * i.e. `extensions/local/authorname/packagename/`
  * PHP files in `src/` (`extensions/local/authorname/packagename/src/`)
  * `config.yml.dist` files in `config/` (`extensions/local/authorname/packagename/config/`)
  * etc

That way it maps to other extension documentation, and requires less "this section is only for …"

Well, that's my thought process anyway, what do you think about all that ~~ramble~~?

bobdenotter: I don't mind sticking with that. Would it require much changes in the code for Bolt to pick them up?

CarsonF:
I wouldn't even include the `index.php` example. No one should be doing that.

Also the `.bolt.php` example is confusing. It should be:
```php
<?php

return [
    'extensions' => [
        new MyExtension\LocalExtensionOne(),
    ]
];
```

or if you are creating the app yourself:
```php
<?php

$app = new Application(); // create or get somehow

$app['extensions'] = $app->share($app->extend('extensions', function ($extensions) {
    $extensions->add(new MyExtension\LocalExtensionOne());

    return $extensions;
}));

return $app;
```

bobdenotter:
@CarsonF Consider it done.

docs/howto/installing-local-extensions.md
```diff
@@ -43,34 +105,17 @@ like this:
...
 
-Bolt will load any extensions that are defined inside the `bolt.yml` file in the
-root of your project. For example you can edit the file to look something like
-this:
+Bolt will load any extensions that are defined inside the `.bolt.yml` file in
+the root of your project. For example you can edit the file to look something
+like this:
 
 ```yaml
 #bolt.yml
```
CarsonF: Let's be consistent with our names: `.bolt.yml`

docs/howto/installing-local-extensions.md
```diff
@@ -80,21 +125,70 @@ paths:
...
+    - Listing\ListingExtension
 ```
 
-Note that each class defined as above eg: `MyextOne\LocalExtensionOne` must have been
-setup using one of the methods defined above.
+Note that each class defined as above eg: `MyextOne\LocalExtensionOne` must
+have been setup using one of the methods defined above.
+
+If you use a PHP autoloader instead of the yaml style one above then you can
+add your extensions in PHP code in either `.bolt.php` or in your own bootstrap.
```
CarsonF: I would remove the paragraph. It is confusing / wrong.

docs/howto/installing-local-extensions.md
```diff
@@ -80,21 +125,70 @@ paths:
...
+        …
+    ]
+    'extensions' => [
+        new MyExtension\LocalExtensionOne(),
+        new Listing\ListingExtension()
+    ]
+];
+```
+
+Example in your own bootstrap:
```
CarsonF: Can we somehow note that this example is also in `.bolt.php`?

docs/howto/installing-local-extensions.md
```diff
@@ -80,21 +125,70 @@ paths:
...
+
+If you are working on a project, and are using one or more local extensions,
+make sure that they are included in your versioning system. Related to that,
+when you're deploying you must make sure the local extension gets deployed as
+well.
+
+Alternatively, the extension can be tracked in its own version control
+repository. In this case it might need to be included separately in the deploy
+process as well. If the local extension is just added to the git repository for
+the main project, this is probably not a big concern.
```
CarsonF: You mean like with regular extensions? 😉

bobdenotter: Yes, basically :-)

docs/howto/installing-local-extensions.md
bobdenotter: Wrong how? This came from Ross' initial draft.

docs/howto/installing-local-extensions.md
bobdenotter: But, that was the example for your own bootstrap, right. Not necessarily in `.bolt.php`.

docs/howto/installing-local-extensions.md
CarsonF: > PHP autoloader

Everyone uses a PHP autoloader 😂 . It's basically just saying you can add extensions in `.bolt.php` if you use that instead of `.bolt.yml`. Which makes it redundant with the examples below.

docs/howto/installing-local-extensions.md
CarsonF: Well if you do a custom bootstrap and don't use `.bolt.php`., then `vendor/bin/nut` won't work. If people are using custom bootstraps, then they know how to extend a pimple service. If they don't then they shouldn't be using a custom bootstrap.

bobdenotter:
Should be GTG now, I think.

GawainLynch:
Sending in… improvements can be follow ups.

PR #648, opened 2017-02-06
Update updating.md (add link to permissions story in update instructions)

evertalbers:
Related an issue I had twice after updating from 3.1 to 3.2 with /app/cache/production not having the correct permissions.
https://boltcms.slack.com/archives/general/p1486394131011459 

Not sure if this deserves a `p class="note"` - too many notes might become chaotic.


GawainLynch:
Thanks mate :+1:

PR #649, opened 2017-02-06
H1 should always be followed by H2. Never directly by H3.

bobdenotter:


GawainLynch:
Then shouldn't our heading layout be adjusted and the CSS too?

image

GawainLynch:
Haha! That'll teach me to try and make sense of anything at 4 in the morning!

Usual disclaimer, I am stepping into visuals here and totally out of my league, so just hopefully passing useful commentary.

My concern (and OCD trigger, but can be ignored as "visuals") is that they are section sub-headings, but H2 is a section heading. See H2 "Example usage in templates" vs H3 "Basic Configuration".

The illustration I am _trying_ to draw (and we know it's not my week for communicating well either), is that you are 100% correct and the cascade should be `h1` -> `h2` -> `h3` etc, but we're forcefully moving sub-section headings to section heading layouts, and IMHO are to big & bold.

Feel free to just ignore, and I'll merge … it's my combination of "being helpful" and "OCD" :smile:

bobdenotter:
If that's the case, we'll need to reconsider the sizes of the headings and such. Keeping the semantic structure is important though, because it also influences the search results.

GawainLynch:
> because it also influences the search results.

That comment alone is enough to "shut me up" :laughing:

PR #656, opened 2017-02-15
hypen cannot be used in contentypes.yml file

forethoughtde:
Change hypen to underscore because hypen is not allowed in contenttypes.yml


SahAssar:
Thanks! 👍

PR #664, opened 2017-02-23
Update controllers-routes.md

diggersworld:
Gives an example of the URL structure which needs to be used to access the callback, as it wasn't obvious that the controller registration played part in the match argument.

### DIFF
--- a/docs/extensions/intermediate/controllers-routes.md
+++ b/docs/extensions/intermediate/controllers-routes.md
@@ -206,3 +206,12 @@ class DropBearController implements ControllerProviderInterface
 }
 ```
 
+The route will then then be accessed via `/dropbear/koala/{type}` for frontend
+controllers and `/{admin_path}/dropbear/koala/{type}` for backend controllers.
+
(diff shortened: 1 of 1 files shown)


docs/extensions/intermediate/controllers-routes.md
```diff
@@ -206,3 +206,5 @@ class DropBearController implements ControllerProviderInterface
 }
 ```
 
+The callback method can then be accessed via `/dropbear/koala/{type}` (frontend) and `/bolt/dropbear/koala/{type}` (backend).
```
SahAssar: IMO it should be "`/dropbear/koala/{type}` (for frontend controllers) or `/bolt/dropbear/koala/{type}` (for backend controllers)".

diggersworld: Updated based on your feedback.

SahAssar: Thanks! 👍

GawainLynch:
@diggersworld great to see you, it's been ages! Let me tweak the branches on this and I'll send it in. :+1:

PR #673, opened 2017-02-25
Update menus.md

opichon:
Original docs were a little confusing in that they did not make it very clear that in order to override the default menu template, a template with the same file name had to be created in the root folder of the theme. This in turn is a little inconsistent with the standard directory layout of themes, where all partials are under the `partials` directory. Not a huge deal, but small inconsistencies create confusion. I have rewritten the code samples by adding the `partials/` prefix to all templates.

I think smart people will readily understand that if they want to put their menu template in the theme root dir, they can; and for dumb people like me, it will save them a few minutes of confusion.


bobdenotter:
Looks good, thank you! 👍

PR #674, opened 2017-02-25
Local extension migration guide

rossriley:
This upgrade guide should cover the move of local extensions during the upgrade to 3.3.

There's one bit still to cover which is the asset sync, but we need to have a further discussion on what strategy we are going to recommend for this.

### DIFF
--- a/docs/extensions/bundled/activation.md
+++ b/docs/extensions/bundled/activation.md
@@ -1,0 +1,43 @@
+---
+title: Activation
+level: advanced
+---
+Activation
+==========
+
+Once your Bundle can be autoloaded, then one more step is needed to enable Bolt
+to load them.
+
+To active a Bundle, you need to add an `extensions` key to either your
+`.bolt.yml` or `.bolt.php` file in the root of your project, with the values
+being the Bundles you want Bolt to load.
+
+
+Updating `.bolt.yml` or `.bolt.php`
+-----------------------------------
+
+An example using `.bolt.yml`:
+
+```yaml
+extensions:
+    - BundleBaseNamespace\MyBundleExtension
+```
+
+To clarify, the value you put in the yml file is exactly what you would use to
+instantiate the class, so in code the above is equivalent to
+`new \BundleBaseNamespace\MyBundleExtension()`
+
+An example using `.bolt.php`
+
+```php
+<?php
+
+use BundleBaseNamespace\MyBundleExtension;
+
+return [
+    'extensions' => [
+        new MyBundleExtension()
+    ]
+];
+```
+
--- a/docs/extensions/bundled/file-location.md
+++ b/docs/extensions/bundled/file-location.md
@@ -1,0 +1,41 @@
+---
+title: File Location & Layout
+level: advanced
+---
+File Location & Layout
+======================
+
+As Bundles are just a part of your site project's code, their location is
+relative to your site's root directory is flexible, and your decision on
+location should be influenced by the necessary file type(s).
+
+
+### Choosing a location
+
+#### PHP only
+
+If your Bundle contains only PHP file(s), and Twig and web assets are kept with
+the site's theme, we recommend using the base namespace of your Bundle as a
+sub-directory of `{site root}/src`,
+
+For example, if your Bundle namespace was `Acme\BundleBaseNamespace`:
+
+```
+{site root}/src/BundleBaseNamespace
+```
+
+
+#### PHP with resources
+
+If you want you Bundle to include non-PHP files, you're better off putting the
+contents into a sub-directory of `extension/` with the pattern
+`extension/{bundle name}`, for example:
+
+| Directory  | Description                                                    |
+| ---------- | -------------------------------------------------------------- |
+| `extension/{bundle name}/src`       | Extension's PHP code
+| `extension/{bundle name}/config`    | Configuration base files (`config.yml.dist`)
+| `extension/{bundle name}/templates` | Twig templates. These should be kept separate from public assets
+
+
+
--- a/docs/extensions/bundled/introduction.md
+++ b/docs/extensions/bundled/introduction.md
@@ -1,0 +1,122 @@
+---
+title: Introduction
+level: advanced
+---
+Introduction
+============
+
+Bundled extensions, or "Bundles", should be thought of as _components_ that are
+_specific_ to **your site project(s)**. As such, developed, managed, and
+deployed using the same tools and methodologies you would deploy your Bolt site
+would normally.
+
+Writing a Bundle has a lot in common with writing a regular extension, such as
+you'd find, and install, from the [Market Place][market]. In that, as simple as
+creating a class in your Application that implements
+`Bolt\Extension\ExtensionInterface`.
+
+However, it is very important to note that, Bundles **do not**:
+  * Manage updating of web included web assets with public directories
+  * Require a `composer.json` file
+  * Auto-configure autoloading
+
+<p class="note"><strong>Note:</strong> Bundles are, by design, intended
+as a collection of <strong>site-specific</strong>, functionality
+enhancing PHP code. File types such as Twig templates, CSS, and
+JavaScript, are better suited being located inside your site's theme
+directory.</p>
+
+
+### Getting Started
+
+The initial design & build of a Bundle has the following aspects that we
+highlight below:
+  * [Bundle files location & layout](#bundle-files-location-amp-layout)
+  * [Bundle loader class design](#bundle-loader-class-design)
+  * [Bundle namespace](#bundle-namespace)
+  * [Building the Bundle](#building-the-bundle)
+  * [Autoloading configuration](#autoloading-configuration)
+  * [Activation configuration](#activation-configuration)
+
+
+#### Bundle namespace
+
+Bundles need to defined inside their own base namespace. This namespace
+is important to decide on first, as it will influence the target
+directory location of your Bundle, and is needed for your _autoloading_,
+and _activation_ configuration
+
+<p class="note"><strong>Note:</strong> The last name in the namespace
+path internally serves as the "author" name.</p>
+
+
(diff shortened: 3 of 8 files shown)


bobdenotter:
Cool. Will proofread and try it out on some old-style local extensions tomorrow

docs/upgrading/migrating-local-extensions-33.md
```diff
@@ -0,0 +1,138 @@
...
+decide to move extensions to be part of your main App namespace and if so you
+can do something like the following: 
+
+```json
+    "autoload" : {
+        "psr-4" : {
+            "MyCompany\\MyApp\\": "src"
+        }
+    }
+```
```
CarsonF: Should we just tell them to use the main `composer.json` in the first place? Less options == less confusion.

docs/upgrading/migrating-local-extensions-33.md
```diff
@@ -0,0 +1,138 @@
...
+then it will look something like this:
+
+```yaml
+
+paths:
+    web: public
+    themebase: public/theme
+    files: public/files
+    view: public/bolt-public/view
+```
```
CarsonF: Do we need to give an example of what the "current" `.bolt.yml` is? My thought is these paths listed are out of date and not in sync with other places in the docs.

docs/upgrading/migrating-local-extensions-33.md
```diff
@@ -0,0 +1,138 @@
...
+$app = new Bolt\Application(array('resources'=>$configuration));
+// Any other application setup carries on here…
+
+// Adding your extension to Bolt
+$app['extensions']->add(new Bolt\Extension\Myname\MyExtensionNamespace\ExtensionClass());
+
+$app->initialize();
+
+return $app;
+```
```
CarsonF: This setup is completely deprecated. `$app->initialize()` doesn't do anything anymore. `extensions` service is called directly. ResourceManager is used instead of PathResolver.

Honestly, I think the PHP version of the bootstrap is for advanced use (at least returning an app instance) and should be minimally documented. Those that need it know how to use it.

docs/upgrading/migrating-local-extensions-33.md
```diff
@@ -0,0 +1,138 @@
...
+
+As you may have realised, the `composer.json` files are now no longer needed for
+local extensions writing a Bolt extension is as simple as creating a class in
+your Application that implements  `Bolt\Extension\ExtensionInterface` and adding
+the Fully Qualified Class Name to your `.bolt.yml` file.
+
+Once you have a local extension loaded in your application, the interface within
+the Bolt extensions screen has also been adjusted to separate local extensions
+from those installed via the Market Place. These local extensions will now
+appear underneath the other installed extensions.
```
CarsonF: This provides a lot of reasoning for the change. IMO this should be moved to the top to explain ourselves.

docs/upgrading/migrating-local-extensions-33.md
rossriley: bear in mind that this is to help people migrating from 3.2 to 3.3, PathResolver hasn't even been included in one release yet, so there's no point documenting it in the migration guide since no-one will be using the code in production.

there's definitely quite a lot of people using a custom bootstrap in production, so it makes sense to cover it off in the migration guide. Once everyone has moved onto the new system then this page can be phased out, it's not intended to stick around forever.

docs/upgrading/migrating-local-extensions-33.md
rossriley: It's identical to this one: https://github.com/bolt/composer-install/blob/release/3.2/.bolt.yml

docs/upgrading/migrating-local-extensions-33.md
rossriley: That is what I thought initially, but @bobdenotter suggested it may be less of a leap to use the extensions one. Personally I'd prefer pointing to the main composer.json since it keeps the extensions one more of a sandbox for us to adjust as needed. 

Ideally it meshes with what I'd hoped too, that people would stop thinking of extensions as a magic black box and more as normal classes that they can load into Bolt to adjust an app as needed.

CarsonF: > Personally I'd prefer pointing to the main composer.json since it keeps the extensions one more of a sandbox for us to adjust as needed.

I agree 100%

> @bobdenotter suggested it may be less of a leap to use the extensions one.

I think it is a necessary leap, and a very small one at that. There's no reason to half way use `extensions/composer.json` and half way not. That causes confusion for both users and us devs.

Keep `extensions/*` a black box for those that want to just install extensions from the UI, that's fine. But anyone loading / creating extensions themselves can use the main `composer.json` for autoloading and add the extension to the bootstrap list themselves. This also prevents NIV here syndrome, as autoloading is the same everywhere.

docs/upgrading/migrating-local-extensions-33.md
CarsonF: Well as of 3.3 it's old. I realize composer-install repo may be out of date. Composer installs don't even need the paths section if they are using default paths.

docs/upgrading/migrating-local-extensions-33.md
GawainLynch: > PathResolver hasn't even been included in one release yet, so there's no point documenting it in the migration guide since no-one will be using the code in production.

I'm calling flawed logic on that one. :disappointed: That would be analogous to us having documented the few legacy extension helpers that survived (should have been dropped anyway) instead of the newer better faster awesome that was a v3 extension.

This iteration of local extensions is "new", and if people are migrating sites, they should be doing that with   forward thinking approaches in mind, it will make their and our support lives easier.

Once we close the door on 3.2, we should make every effort to document/advise/etc use of the newer systems and tell epople to remove/replace any use of RM in existing code.

docs/upgrading/migrating-local-extensions-33.md
GawainLynch: I've been (sometimes gumpily) all over the place on this one since @CarsonF first started bugging me about it on one of my PRs.

In an attempt to prove my point, I looked over the docs and notices that we've had, or are developing, a habit of re-documenting tiny bits in many places leading to inconsistency in the docs.

To that end, my vote w.r.t. approach on these is to **only** show the change, but put a hyperlinked note near said blocks to the "canonical" documentation for the surrounding sections like `paths:` … strikes me as making all of our lives easier in the long-run with keeping our documentation consistent.

GawainLynch:
Wait! :laughing: 

I have a tonne coming, I was simply asking in Slack if we collectively like the approach … I have a lot more I want to move around and push (today?).

docs/extensions/bundled/activation.md
```diff
@@ -0,0 +1,43 @@
+---
+title: Activation
+level: advanced
+---
+Activation
+==========
+
+Once your Bundle can be autoloaded then one more step is needed to
```
bobdenotter: There should be a comma before 'then', right? 

I can't english, so i'm not sure. :-)

docs/extensions/bundled/activation.md
```diff
@@ -0,0 +1,43 @@
...
+enable Bolt to load them.
+
+To active an Bundle, you need to add an `extensions` key to either your
```
bobdenotter: "a bundle"

docs/extensions/bundled/activation.md
```diff
@@ -0,0 +1,43 @@
...
+
+An example using `.bolt.yml`:
+
+```yaml
+extensions:
+    - BundleBaseNamespace\MyBundleExtension
+```
+
+To clarify, the value you put in the yml file is exactly what you would use to
+instantiate the  class, so in code the above is equivalent to
```
bobdenotter: double space

docs/extensions/bundled/file-location.md
```diff
@@ -0,0 +1,42 @@
+---
+title: File Location & Layout
+level: advanced
+---
+File Location & Layout
+======================
+
+As Bundles are just a as part of your site project's code, their
```
bobdenotter: "a as"

docs/extensions/bundled/file-location.md
```diff
@@ -0,0 +1,42 @@
...
+location relative to your site's root directory is flexible, and your
+decision on location should be influenced by the necessary file contents
+type(s).
```
bobdenotter: "contents types"? Im not sure what is meant here.

docs/extensions/bundled/introduction.md
```diff
@@ -0,0 +1,122 @@
...
+_specific_ to **your site project(s)**. As such, developed, managed, and
+deployed using the same tools and methodologies you would deploy your Bolt site
+would normally.
+
+Writing a Bundle has a lot in common with writing a regular extension, such as
+you'd find, and install, from the [Market Place][market]. In that, as simple as
+creating a class in your Application that implements
+`Bolt\Extension\ExtensionInterface`.
+
+However, it is very important to note that, Bundles **do not**:
```
bobdenotter: "that, Bundles" -> i think that comma should be removed.

docs/extensions/bundled/introduction.md
```diff
@@ -0,0 +1,122 @@
...
+extensions.
+
+[Activation section][activation]
+
+
+Difference to Market Place Extensions
+-------------------------------------
+
+<p class="note"><strong>Note:</strong> One big difference between bundled and
+normal sextensions, is that bundled extensions <strong>do not</strong> require
```
bobdenotter: sextensions. 🤣

docs/extensions/bundled/introduction.md
GawainLynch: No, that one is correct. The point is "However Bundles", what is in between is an interjection.

bobdenotter: Ah, yes.. You are correct. :-)

PR #675, opened 2017-02-25
Update installing-local-extensions.md

Ichag:
Remove webroot from local extensions path

### DIFF
--- a/docs/howto/installing-local-extensions.md
+++ b/docs/howto/installing-local-extensions.md
@@ -10,7 +10,7 @@ extensions site.
 
 There are a couple of caveats:
 
-- Must be located in `{web_root}/extensions/local/{author_name}/{extension_name}/`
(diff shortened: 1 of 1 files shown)


docs/howto/installing-local-extensions.md
```diff
@@ -10,7 +10,7 @@ extensions site.
 
 There are a couple of caveats:
 
-- Must be located in `{web_root}/extensions/local/{author_name}/{extension_name}/`
```
GawainLynch: That is incorrect based on the rest of the document. In hindsight with the changes in 3.3 `{site_root}` is a better choice, but there is also the directories `{site_root}/public/extensions/ …` where the `public/` part if far more appropriate to call "web root" :smile:

GawainLynch:
Updated for you, thanks :+1:

PR #678, opened 2017-02-27
removed duplicate float field type

larsvansche:
removed duplicate float field type between lines 224 and 228


bobdenotter:
Merged into `release/3.0`. Thank you

PR #687, opened 2017-03-13
Docs permissions flat structure

sablonier:
Just a small proposal to have separated content in docs->install->permissions for flat structure installs.


bobdenotter:
Thanks, @sablonier. We'll merge this in, in a bit! 👍

PR #698, opened 2017-03-29
Example code gave error in strict variables mode

anketwokings:
In the submenu, one of the items had no path or link. This results in an error in the default (base 2016) bolt menu twig.


anketwokings:
Thanx!

PR #699, opened 2017-03-30
Twig functions filters tests tags

GawainLynch:
* Document `{{ url() }}`
* Update `{{  path() }}`
* Separate tags into tags
* Separate filters into filters
* Separate functions into functions


bobdenotter:
Sweet! I spotted some typo's, but we can fix those. 

I think we should also better describe the differences between `url`, `path` and `asset`, and when you'd want to use one or the other. 
Should these three perhaps get their own page with the detailed description of the three, and just a small blurb in the list with a link?

GawainLynch:
> Should these three perhaps get their own page

That's what I was doing a bad job of trying to say needed to happen on Slack this morning … :rofl: 

> I think we should also better describe the differences between `url`, `path` and `asset`, and when you'd want to use one or the other.

Yes, but can we do a follow up PR?

> I spotted some typo's

Shall rebase magic :koala:

GawainLynch:
>> Should these three perhaps get their own page

> That's what I was doing a bad job of trying to say needed to happen on Slack this morning

It would also make it easier to have a common section in there for `url` and `path` on common route binding names.

PR #719, opened 2017-04-22
Add clearfix div, to prevent flow issues.

bobdenotter:
As reported by @harikt on twitter. Thanks! 👍 

https://twitter.com/harikt/status/855477515426705411

### DIFF
--- a/view/templates/_master.twig
+++ b/view/templates/_master.twig
@@ -57,6 +57,8 @@
         </ul>
     </nav>
 
+    <div class="clearfix"></div>
+
     <main class="row">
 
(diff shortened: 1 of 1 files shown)


GawainLynch:
Deployed too.

view/templates/_master.twig
```diff
@@ -57,6 +57,8 @@
         </ul>
     </nav>
 
+    <div class="clearfix"></div>
```
harikt: Oh the template was here :) . I thought only markdown was here.

Thank you.

PR #722, opened 2017-04-23
fix wrong documentation, add missing documentation

serweb-labs:


GawainLynch:
LGTM … passing to Ross to confirm.

Moved to `release/3.1` as that is where this first appears, and we'll cascade from there.

PR #728, opened 2017-05-01
Performance section

GawainLynch:
This starts the breakout of the HOWTO into a performance section, most notably to include the HTTP Cache reverse proxy section that I'll PR when this cascades to `release/3.2`


GawainLynch:
Redirect in-place on the web server.

PR #740, opened 2017-05-12
Bolt 3.3+ Amazon S3 Remote Filesystem

GawainLynch:
These are the starter docs for using Amazon S3 as a remote filesystem in Bolt :tada: 

I would like to point out that a huge amount of work on the Bolt & Filesystem repositories, and most of the example code here is thanks to @CarsonF … seriously, if you knew the hours that have gone into making this happen! :smile: :+1:

### DIFF
--- a/docs/performance/remote-filesystem/amazon-s3.md
+++ b/docs/performance/remote-filesystem/amazon-s3.md
@@ -1,0 +1,354 @@
+---
+title: Amazon S3
+level: advanced
+---
+Amazon S3
+=========
+
+Configuring Bolt to use Amazon Simple Storage Service (Amazon S3) requires the
+addition of Flysystem and caching libraries, as well as custom service provider
+code to your project.
+
+1. Set-up, configure, and confirm operational your Amazon S3 bucket
+1. Update `composer.json` in your project root
+1. Creating service providers
+1. Bootstrap your providers
+
+Prerequisites
+-------------
+
+<p class="warning"><strong>Warning:</strong> This documentation, and examples,
+assume you have a fully configured and working Amazon S3 bucket.
+</p>
+
+
+
+### Required Libraries
+----------------------
+
+You need to add the following libraries to your project:
+
+  * `league/flysystem-aws-s3-v3:^1.0`
+  * `league/flysystem-cached-adapter:^1.0`
+
+This can simply be achieved by running:
+
+```bash
+$ composer require league/flysystem-aws-s3-v3:^1.0 league/flysystem-cached-adapter:^1.0
+```
+
+
+### Caching Libraries
+----------------------
+
+While other cache targets are optional, right now this example will be based on
+Redis. Alternatively any PSR-6 can be used, as well as Doctrine Cache.
+
+To use Redis, you will need a configured and running Redis instance, and need
+to install `predis/predis:^1.1`, which can be achieved by running:
+
+
+```bash
+$ composer require predis/predis:^1.1
+```
+
+The cache needs to be shared between servers caching for the same site, else
+cache results will be inconsistent, unreliable, and self-defeating of the aim
+of caching.
+
+We recommend using [Amazon ElastiCache][elasticache] for a shared Redis
+instance, and using the Predis library to connect to it.
+
+
+Autoloading
+------------
+
+Autoloading of your service providers can be set-up in the same way as
+[Extension Bundles][bundles], by adding your base namespace and source
+directory ot the `psr-4` section of your project root `composer.json` file,
+e.g.
+
+```json
+    "autoload": {
+        "psr-4": {
+            "MySiteApp\\": "src"
+        }
+    }
+```
+
+
+<p class="note"><strong>Note:</strong> If you change anything in the
+<code>psr-4</code> section of your <code>composer.json</code> file, remember to
+re-run <code>composer dump-autoload</code> to update the cached autoloading
+files.
+</p>
+
+
+Service Providers
+-----------------
+
+### Amazon AWS
+
+The provider for Amazon AWS services needs to define `app['aws.s3']`, that
+returns an `\Aws\S3\S3Client` used to interact with Amazon Simple Storage
+Service (Amazon S3).
+
+For detailed information on constructing these classes, and the options
+provided see the [Amazon S3 documentation][s3-docs], and specifically the
+[S3 SDK documentation][s3-sdk] and [S3 ApiGen documentation][s3-api].
+
+
+#### Amazon Credentials
+
+This credential provider first checks in order:
+  * Amazon `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables
+  * "default" profile in ~/.aws/credentials, or the profile specified by the
+    `AWS_PROFILE` environment variable
+  * "profile default" profile in ~/.aws/config (which is the default profile of
+    AWS CLI)
+  * Fetch credentials from the ECS environment variables
+  * Fetches credentials for the EC2 instance's IAM Role
+    * If you are using EC2 instances then this is the recommended way to
+      provide credentials
+
+For information on the `Aws\Credentials` classes, see Amazon's ApiGen
+documentation for [AWS Credentials][s3-api-creds].
+
+If using an IAM policy, you'll need to attach the follwoing to the IAM user
+role:
+
+```json
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Effect": "Allow",
+            "Action": [
+                "s3:HeadBucket",
+                "s3:GetBucketAcl"
+                "s3:HeadObject",
+                "s3:GetObject",
+                "s3:GetObjectAcl",
+                "s3:GetObjectVersion",
+                "s3:GetObjectVersionAcl",
+                "s3:PutObject",
+                "s3:PutObjectAcl",
+                "s3:PutObjectVersionAcl",
+                "s3:DeleteObject",
+                "s3:DeleteObjectVersion"
+            ],
+            "Resource": "arn:aws:s3:::bucket-name/*"
+        }
+    ]
+}
+```
+
+
+### Filesystem
+
+#### Filesystem Cache Factory
+
+Create a protected callable `$app['filesystem.cache_factory']` that when
+invoked returns an instance of `\Bolt\Filesystem\Adapter\Cached`.
+
+
+#### S3 Filesystem Service
+
+Create a protected callable `$app['filesystem.s3']` that when invoked will
+return an a class implementing `Bolt\Filesystem\FilesystemInterface`.
+
+If using `\Bolt\Filesystem\Filesystem` as your `FilesystemInterface` object,
+you first need to construct an `\Bolt\Filesystem\Adapter\S3` adapter class.
+
+Your adapter can then be passed to the `$app['filesystem.cache_factory']`
+callable, and the return value used as the adpater parameter to construct
+`Filesystem`.
+
+
+#### Extend Filesystem
+
+Next, you must extend `$app['filesystem']`, and call `mountFilesystems()` using
+an associative array of the filesystem(s) you want to override using the S3
+filesystem callable you define.
+
+
+### Assets
+
+In order for functionality, like Twig's `{{ asset() }}` function to know about
+the variation in location of assets that is return to the requesting client,
+you need to next override the desired package in the package group with a
+`Symfony\Component\Asset\UrlPackage` object that defines your base Amazon S3
+bucket URL.
+
+This is achived by extending `$app['asset.packages']`, and adding your specific
+`UrlPackage` object to he desired package pool name.
+
+
+### Example Provider
+
+This is an example of the `src/Provider/AwsServiceProvider.php` file.
+
+```php
+<?php
+
+namespace MySiteApp\Provider;
+
+use Aws\Credentials\CredentialProvider;
+use Aws\Credentials\Credentials;
+use Aws\S3\S3Client;
+use Bolt\Filesystem\Adapter;
+use Bolt\Filesystem\Adapter\Cached;
+use Bolt\Filesystem\Cached\DoctrineCache;
+use Bolt\Filesystem\Filesystem;
+use Doctrine\Common\Cache\PredisCache;
+use Predis;
+use Silex\Application;
+use Silex\ServiceProviderInterface;
+use Symfony\Component\Asset\UrlPackage;
+
+class AwsServiceProvider implements ServiceProviderInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function register(Application $app)
+    {
+        /*
+         * Amazon S3 provider.
+         */
+        $app['aws.s3'] = $app->share(function ($app) {
+            return new S3Client([
+                // The bucket's region name, e.g. 'eu-central-1'
+                'region' => 'my-bucket-location',
+                'version' => '2006-03-01',
+            ]);
+        });
+
+        /*
+         * Filesystem.
+         */
+        $app['filesystem'] = $app->share($app->extend(
+            'filesystem',
+            function ($filesystem, $app) {
+                $bucketName = 'your S3 bucket name';
+                $prefix = '';
+
+                $filesystems = [
+                    'files' => $app['filesystem.s3']('files', $bucketName, $prefix)
+                ];
+                $filesystem->mountFilesystems($filesystems);
+
+                return $filesystem;
+            }
+        ));
+        // The S3 filesystem protected callable.
+        $app['filesystem.s3'] = $app->protect(
+            function ($name, $bucket, $prefix = '') use ($app) {
+                $adapter = new Adapter\S3($app['aws.s3'], $bucket, $prefix);
+                $adapter = $app['filesystem.cache_factory']($adapter, 'flysystem-' . $name, 3600);
+
+                return new Filesystem($adapter);
+            })
+        ;
+        // Cache factory
+        $app['filesystem.cache_factory'] = $app->protect(
+            function ($adapter, $name, $expire = null) {
+                // Note: Predis is use here as an example
+                $cache = new DoctrineCache(new PredisCache(new Predis\Client()));
+
+                return new Cached($adapter, $cache, $name, $expire);
+            }
+        );
+
+        /*
+         * Assets.
+         */
+        $app['asset.packages'] = $app->share($app->extend(
+            'asset.packages',
+            function ($packages, $app) {
+                $package = new UrlPackage(
+                    // CloudFront CDN url can also be used here
+                    'http://example.s3-website.eu-central-1.amazonaws.com',
+                    $app['asset.version_strategy']('files'),
+                    $app['asset.context']
+                );
+                $packages->addPackage('files', $package);
+
+                return $packages;
+            }
+        ));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function boot(Application $app)
+    {
+    }
+}
+```
+
+
+A note on the parameters passed to `UrlPackage`:
+  - First parameter for `UrlPackage` is the base URL or your Amazon S3 bucket
+  - Second parameter is the version strategy applicable to the package
+  - Third parameter is the **must** be the name of the package you're
+    replacing, and **must** be the mount point of the filesystem you are
+    changing. This also applies to the first parameter used in `addPackage()`
+    as well.
+
+In this example, 'files' is package name, that is package that is responsible
+for the contents of your project's `files/` mount point.
+
+Optionally, you may with to construct your own `\Aws\Credentials\Credentials`
+object, and pass that to `CredentialProvider::fromCredentials()`, e.g.
+
+```php
+    $app['aws.s3'] = $app->share(function ($app) {
+        $accessKey = 'your access key from Amazon';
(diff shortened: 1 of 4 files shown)


docs/performance/remote-filesystem/introduction.md
```diff
@@ -0,0 +1,14 @@
...
+---
+Introduction
+============
+
+On more complex, and high-volumne sites, using Content Delivery Networks (CDN)
+can offer better loading times of file assets due to them being mirrored in
+geographically relevant data centres around the world.
+
+To support this seamlessly in the Bolt application, our filesystem service
+provides support for Remote Filesystem adapters individicual filesystems.
```
CarsonF: typo.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+
+  * `league/flysystem-aws-s3-v3:^1.0`
+  * `league/flysystem-cached-adapter:^1.0`
+  * `league/flysystem-memory:^1.0`
+
+This can simply be achieved by running:
+
+```bash
+$ composer require league/flysystem-aws-s3-v3:^1.0 league/flysystem-cached-adapter:^1.0 league/flysystem-memory:^1.0
+```
```
CarsonF: `flysystem-memory` isn't required.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+
+
+### Caching Libraries
+----------------------
+
+While other cache targets are optional, right now this example will be base on Redis.
```
CarsonF: Maybe say that any PSR-6 or Doctrine Cache instances can be used?

Also there's a reason I used Redis. The cache needs to be shared amongst all the nodes. We are using Elasticache for this.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+        }
+    }
+```
+
+
+<p class="note"><strong>Note:</strong> If you change anything in the
+<code>psr-4</code> section of your <code>composer.json</code> file, remember to
+re-run <code>composer dump-autoload</code> to update the cached autoloading
+files.
+</p>
```
CarsonF: This is "advanced" but we are still going over autoloading?...

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+        $app['aws'] = $app->share(
+            function () {
+                $options = [
+                    // The bucket's region name, e.g. 'eu-central-1'
+                    'region' => 'my-bucket-location',
+                ];
+
+                return new Sdk($options);
+            }
+        );
```
CarsonF: Let's skip the `Sdk` part. S3 clients can be created directly - `new S3Client`

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+
+For detailed information on constructing these classes, and the options
+provided see the [Amazon S3 documentation][s3-docs], and specifically the
+[S3 SDK documentation][s3-sdk] and [S3 ApiGen documentation][s3-api].
+
+
+#### Amazon Credentials
+
+In this example we use `CredentialProvider::defaultProvider()` to get your
+Amazon credentials.
```
CarsonF: I can't think of a case where the defaults wouldn't be wanted since it basically checks everywhere. And if you don't follow their defaults then other tools have to have the creds passed in manually too.

Let's just leave the code out. I think it's still good to list where the defaults check.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+
+This credential provider first checks in order:
+  * Amazon specific environment variables
```
CarsonF: `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+  * "default" profile in ~/.aws/credentials
```
CarsonF: Or the profile specified by `AWS_PROFILE` env var.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+  * "profile default" profile in ~/.aws/config (which is the default profile of
+    AWS CLI)
+  * Tries to make a GET Request to fetch credentials if the Ecs environment
+    variable is presented
```
CarsonF: Simplify to `ECS environment variables`

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+  * Checks for EC2 instance profile credentials
```
CarsonF: Change to "Fetches credentials for the EC2 instance's IAM Role".

Would also note that "If you are using EC2 instances then this is the recommended way to provide credentials".

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+        );
+
+        $app['aws.s3'] = $app->share(function ($app) {
+            $credentials = CredentialProvider::defaultProvider();
+            $options = [
+                'version' => '2006-03-01',
+                'credentials' => $credentials,
+            ];
+
+            return $app['aws']->createS3($options);
```
CarsonF: to reiterate
```php
return new S3Client([
  'region' => 'us-east-1', // bucket's region
  'version' => '2006-03-01',
]);
```

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+        });
+```
+
+<p class="note"><strong>Note:</strong> This particular exacmple should be
+considered conceptual, to simply illustrate the basic requirements.
+<br>
+Saving of thngs like secrect keys to code that gets committed to version
+control is bad security practice.
+</p>
```
CarsonF: Nah, drop this section. K.I.S.S.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+    }
+}
+```
+
+
+The first parameter for `UrlPackage` is the base URL or your Amazon S3 bucket.
+
+The second parameter is the version strategy applicable to the package, the
+value in the final parentheses **must** be the name of the package you're
+replacing.
```
CarsonF: It must be the name of the filesystem you are replacing. But it should match the package name as well, or the fs asset plugin will break.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+
+In this example, 'files' is package name, that is package that is responsible
+for the contents of `{web root}/files/`.
```
CarsonF: This is confusing, as it's not a local directory anymore.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+
+<p class="note"><strong>Note:</strong> This particular exacmple should be
+considered conceptual, to simply illustrate the basic requirements.
+<br>
+Saving of thngs like secrect keys to code that gets committed to version
+control is bad security practice.
+</p>
+
+
+### Assets
```
CarsonF: I would move this below the filesystem section. Filesystem is the next logical jump from AWS configuration.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+     * {@inheritdoc}
+     */
+    public function register(Application $app)
+    {
+        $app['asset.packages'] = $app->share(
+            $app->extend(
+                'asset.packages',
+                function ($packages, $app) {
+                    $package = new UrlPackage(
+                        'http://example.s3-website.eu-central-1.amazonaws.com',
```
CarsonF: CloudFront CDN url can also be used here.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,396 @@
...
+Your adapter can then be passed to the `$app['filesystem.cache_factory']`
+callable, and the return value used as the adpater parameter to construct
+`Filesystem`.
+
+
+#### Extend Filesystem
+
+Finally, you must extend `$app['filesystem']`, and call `mountFilesystems()`
+using an assicative array of the filesystem(s) you want to override using the
+`$app['filesystem.s3']` protected callable.
```
CarsonF: I would flip this order, so the user can follow why. "We want to override the "files" filesystem." -> We need an S3 adapter for that -> The adapter should be cached for performance.

Also you are mixing general documentation with specifics from the example below. Maybe it would be better to have those separated. Or maybe have more comments in the example below?

CarsonF:
Here's the IAM policy that you'll need to attach to the IAM role / user.
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:HeadBucket",
                "s3:GetBucketAcl"
                "s3:HeadObject",
                "s3:GetObject",
                "s3:GetObjectAcl",
                "s3:GetObjectVersion",
                "s3:GetObjectVersionAcl",
                "s3:PutObject",
                "s3:PutObjectAcl",
                "s3:PutObjectVersionAcl",
                "s3:DeleteObject",
                "s3:DeleteObjectVersion"
            ],
            "Resource": "arn:aws:s3:::bucket-name/*"
        }
    ]
}
```

docs/performance/remote-filesystem/introduction.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: I was, and still am, on the fence on this … and increasingly leaning towards _leaving_ it. It doesn't provide _"documentation"_ that is repeated elsewhere, just a _reminder_ for those that are prone to forget these things, i.e. they don't think like Carson. :wink:

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Clarified, and removed `{web root}` also

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Rehashed it, but I'll get you to have a re-read when I push the `asdf` commit

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

GawainLynch:
@CarsonF I have pushed a commit "Review updates" that I'll squash pre-merge … so let me have it :wink:

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,377 @@
...
+        });
+```
+
+
+### Filesystem
+
+#### Filesystem Cache Factory
+
+Create a protected callable `$app['filesystem.cache_factory']` that when
+invoked returns a class extending `\League\Flysystem\Cached\CachedAdapter`.
```
CarsonF: This needs to be our subclass instead of Leagues actually

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+```
+
+
+The first parameter for `UrlPackage` is the base URL or your Amazon S3 bucket.
+
+The second parameter is the version strategy applicable to the package, the
+value in the final parentheses **must** be the name of the package you're
+replacing, and **must** match the package name used in `addPackage()` as well.
```
CarsonF: > **must** be the name of the package you're replacing

**must** be the mount point of the filesystem you are changing.

docs/performance/remote-filesystem/amazon-s3.md
```diff
@@ -0,0 +1,377 @@
...
+While other cache targets are optional, right now this example will be based on
+Redis. Alternatively any PSR-6 can be used, as well as Doctrine Cache.
+
+To use Redis, you will need a configured and running Redis instance, and need
+to install `predis/predis:^1.1`, which can be achieved by running:
+
+
+```bash
+$ composer require predis/predis:^1.1
+```
```
CarsonF: The cache needs to be shared between servers, else bad things happen. We recommend using Elasticache for a shared Redis instance, and using the Predis library to connect to it.

GawainLynch:
> Still missing the IAM example too, unless you think it's not needed.

Yeah, I wanted to put it in, I didn't know what to do with it! 

I have no objections, just haven't had the time to combine reading up Amazon API docs as well :rofl:

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done, with links :+1:

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

docs/performance/remote-filesystem/amazon-s3.md
GawainLynch: Done

GawainLynch:
OK, heavily updated mate.

docs/performance/remote-filesystem/introduction.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Introduction
+---
+Introduction
+============
+
+On more complex, and high-volumne sites, using Content Delivery Networks (CDN)
```
bobdenotter: Typo: `volumne`

docs/performance/remote-filesystem/amazon-s3.md
bobdenotter: I'd say it should stay in. People are very good at just skipping over stuff they don't need, so there's no real harm in keeping it. Alternatively, for a bunch of people this is not trivial. :-)

docs/performance/remote-filesystem/introduction.md
GawainLynch: Fixed

GawainLynch:
> Still missing the IAM example too, unless you think it's not needed.

Had a very quick read up, modified your sentence slightly and added the JSON for the policy.

siada:
Need to add a note about requiring `symfony/browser-kit` as this doesn't work without it

GawainLynch:
> Need to add a note about requiring `symfony/browser-kit` as this doesn't work without it

What requires BrowserKit?

siada:
I'm not 100% sure of the root, but the error itself is thrown by `symfony\http-kernel`

GawainLynch:
I just checked on my local install, that does **not** have BrowserKit installed, it work, and is only being invoked in the whole tree by tests … something you've configured locally?

siada:
Hmm... I thought it might be this as the error only occurs when the AwsServiceProvider is included in .bolt.php

GawainLynch:
How recent is the `bolt/bolt` 3.3-beta you're working with?

siada:
I'm running beta-10

GawainLynch:
OK, so not what I was thinking then. :+1:

CarsonF:
@siada did you update your composer requirements (as mentioned in doc)?

siada:
Yes, I executed:

CarsonF:
Well then did you change AwsServiceProvider to not use Predis then? ;)

siada:
So I just tested it on a completely brand new installation, this is the jist of my process:
`composer create-project bolt/composer-install:^3.2`
then update composer.json to be 3.3-beta10 instead of ^3.2
`composer require league/flysystem-aws-s3-v3:^1.0 league/flysystem-cached-adapter:^1.0`
`mkdir -p src/Provider`
`vi src/Provider/AwsServiceProvider.php`
Paste in example from amazon-s3.md
create .bolt.yml and add the AwsServiceProvider
`composer dumpautoload`
try load homepage -> image

so it looks like somethings not right? I fully understand that I've not done any of the configuration for the S3 bucket, but it's not even getting as far as that as there is a package which is considered 'missing'.

CarsonF:
@siada you didn't require `predis/predis` but the example `AwsServiceProvider` uses it. https://github.com/bolt/docs/pull/740/files#diff-04e49214ef18381423b26544246c2a63R257

siada:
Oh... that did actually get rid of the error 😹 why would installing predis fix an error with http-kernel\Client trying to load browserkit ?

CarsonF:
`Predis\Client` is trying to be loaded from a call made within the BrowserKit namespace. It's then trying to give you a list of Clients, one of which you "forgot" to import. However, the Predis\Client isn't in that list because the class doesn't exist because you never installed it.

ehough:
Just as a point of reference, I was able to follow these instructions about two weeks ago and got my Bolt installation working with S3. I followed the docs carefully and didn't encounter any major hangups along the way. Using PHP 7.1.5, Bolt 3.3 beta, phpredis, and ElastiCache. Really appreciate your great work on this.

GawainLynch:
@siada That is the kind of mistake that I usually make, about twice a day :rofl: … great to know it is sorted, and thank you for testing and giving feedback!

@ehough Thank you for the feedback also!

It is fantastic that the documentation, and our implementation, of the new feature is getting wider testing! :+1: … @CarsonF and I are usually fairly O.C.D. in our approach, but human. 

Thank you all :smile:

siada:
No problem! as @ehough pointed out, it does work very well, once I got it running 😁 I think it would be beneficial at some point to get this into the core of Bolt along with support for other providers.

GawainLynch:
> I think it would be beneficial at some point to get this into the core of Bolt

Well, the code to enable it to run is in core, and will be released as 3.3 stable very soon. :tada: 

But carrying the dependencies is way out of scope for a base install. This has, since the feature development started, been targeted at people who are building their own Composer project and are comfortable doing so. i.e. if one knows enough to set up an S3 bucket, get it working, write the service providers, and get the loading … one can manage a `composer require` without blinking an eye :wink: 

>  along with support for other providers.

Yes, as time/need/PRs permit we'll add more adaptors to `bolt/filesystem` … but small number of coders versus large amount of project work :smile_cat:

siada:
I don't suppose this handles the thumbnail generation does it? I'm trying to keep all images out of my scc repository

CarsonF:
@siada yes it does 😄

GawainLynch:
**Merge note:** This is waiting on a branch push from @CarsonF … :clock10:

bobdenotter:
@CarsonF, @GawainLynch Is this still waiting? I know of a few distinct people who have used this to successfullt implement S3 or even Dropbox Filesystems.

GawainLynch:
@bobdenotter: Yeah, still as-per my comment above yours … @CarsonF has a push pending, I gather he just ran out of time before his break. Let's ping him on Slack this afternoon and see if he objects to an as-is merge.

bobdenotter:
In the spirit of moving forward, I say we merge this one in as-is. As noted above, it's already quite extensive, and multiple people outside of the core team have reviewed it (by de facto using it to implement this feature in a project). 

BDFL says: It's Good Enough™.

PR #744, opened 2017-06-08
Command "dump-autoloader" is not defined.

xiaohutai:
Just sayin':

```
 [Symfony\Component\Console\Exception\CommandNotFoundException]
  Command "dump-autoloader" is not defined.
  Did you mean one of these?
      dumpautoload
      dump-autoload
```


GawainLynch:
Moved to 3.3, but thank you … That one must have been when the Stupid Jar had the lid off :rofl:

PR #746, opened 2017-06-11
Add setPriority for widgets and more sensible defaults for priority

SahAssar:
* setPriority for widgets is already implemented here, this just adds docs: https://github.com/bolt/bolt/blob/master/src/Asset/Widget/Widget.php#L270-L275
* Making the examples for priority 99 seems pretty odd, this changes them to 5 and changes the text to make it more clear that it sets the order, not just pushing a single asset/widget to the bottom.
* Also fixes some long lines in the widget section.

### DIFF
--- a/docs/extensions/basics/web-assets.md
+++ b/docs/extensions/basics/web-assets.md
@@ -23,6 +23,8 @@ subdirectory of `web/`.
 <p class="note"><strong>Note:</strong> Local extensions (often used for testing)
 will not have their assets auto-copied to the web folder, see the <a href="../../howto/installing-local-extensions#step-4">how-to on the subject</a>
  for more info.</p>
+ 
+<p class="note"><strong>Note:</strong> `setPriority` is subject to change in the next major release of bolt.</p>
 
 Registering Assets
 ------------------
--- a/docs/extensions/basics/web-assets.md
+++ b/docs/extensions/basics/web-assets.md
@@ -24,7 +24,7 @@ subdirectory of `web/`.
 will not have their assets auto-copied to the web folder, see the <a href="../../howto/installing-local-extensions#step-4">how-to on the subject</a>
  for more info.</p>
  
-<p class="note"><strong>Note:</strong> `setPriority` is subject to change in the next major release of bolt.</p>
+<p class="note"><strong>Note:</strong> `setPriority` is subject to change in the next major release of Bolt.</p>
 
 Registering Assets
 ------------------
--- a/docs/extensions/basics/web-assets.md
+++ b/docs/extensions/basics/web-assets.md
@@ -95,7 +95,7 @@ protected function registerAssets()
     $asset = new Stylesheet();
     $asset->setFileName('koala.css')
         ->setLate(true)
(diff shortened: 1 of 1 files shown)


docs/extensions/basics/web-assets.md
```diff
@@ -110,8 +110,8 @@ In the above example:
   * `setFileName()` sets the file name
   * `setLate(true)` tells the asset injector to insert the `<link rel="stylesheet" href="path/web/koala.css">`
     at the end of the HTML `<body>`
-  * `setPriority(99)` tells the injector to insert this `<link rel="stylesheet" href="path/web/koala.css">`
-    *after* any other stylesheet tags in that location with a priority lower than 99
+  * `setPriority(5)` tells the injector when to insert the stylesheet. Lower
```
GawainLynch: If it is lower first, we have a bug in core.

SahAssar: Hmm... It's the opposite in core? Do you understand the previous text to imply that too? (I'm asking, since in my mind the previous text also said "lower first", just in a move convoluted way.)

GawainLynch: Oh, I see … it looks like we either have a :bug: in core, or a :koala: :bug: 

Basically, if you have a second/set up to double check on, as with Silex/Symfony event listeners, the higher the number, the sooner it should load/apply, i.e. 1 is a higher priority that -1 

BTW, the change seems good otherwise, and clearer, lets just figure out what is going on here and merge after that.

SahAssar: Ah, right, that makes sense... lemme verify that's how it works currently and then change it.

SahAssar: Lower first is the way it's handled for css, js and widgets currently (I tested all three, @bobdenotter tested widgets and verified it), so these docs seem correct for now at least. Changing it to the other way around makes sense to me, but doing it now might break extensions, no?

GawainLynch: Thanks for checking … lets pick this up on Slack, as there are two big problems; a) currently core's logic is wrong, and the rest of the world does it correctly; b) BC :disappointed:

docs/extensions/basics/web-assets.md
```diff
@@ -23,6 +23,8 @@ subdirectory of `web/`.
 <p class="note"><strong>Note:</strong> Local extensions (often used for testing)
 will not have their assets auto-copied to the web folder, see the <a href="../../howto/installing-local-extensions#step-4">how-to on the subject</a>
  for more info.</p>
+ 
+<p class="note"><strong>Note:</strong> `setPriority` is subject to change in the next major release of bolt.</p>
```
bobdenotter: CapitalBdangit();

bobdenotter:
As discussed in the meeting of 2017-06-13: Merging in!

GawainLynch:
This should have gone into `release/3.0` … let me fix :disappointed:

GawainLynch:
Deployed on all branches now, thanks mate :+1:

PR #749, opened 2017-06-15
Added a new working example to the HTML editor field type

smgdesign:
Example now includes entities and entities_latin with explanation of what they store


bobdenotter:
Thanks for the addition, @smgdesign! 👍

PR #753, opened 2017-06-16
Update Bundled Extension Activation

siada:
Current 3.3 documentation shows that .bolt.php should return an array of new instances of bundled extensions to load. In reality this should actually be an array of fully-qualified class names.

### DIFF
--- a/docs/extensions/bundled/activation.md
+++ b/docs/extensions/bundled/activation.md
@@ -29,7 +29,7 @@ To clarify, the value you put in the yml file is exactly what you would use to
 instantiate the class, so in code the above is equivalent to
 `new \BundleBaseNamespace\MyBundleExtension()`
 
-An example using `.bolt.php`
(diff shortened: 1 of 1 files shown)


GawainLynch:
> In reality this should actually be an array of fully-qualified class names.

That is not correct, [it will accept either](https://github.com/bolt/bolt/blob/release/3.3/src/Bootstrap.php#L162-L168)

Want to add two to make the point clearer with a comment?

siada:
I did try it as a list of objects and I got the following error:

GawainLynch:
Not sure what's going on there … was trying to clear a long over-running TODO off my list and check after that … watch this space! :smile:

GawainLynch:
OK, PR https://github.com/bolt/bolt/pull/6767 landing in a 3.3 branch near you that fixes extensions as objects from `.bolt.php`. Sorry for the delay, been a lot to get done lately :smile: 

Thanks very much :+1:

PR #757, opened 2017-06-21
Mention 'upload' option (4 of 4)

evertalbers:
Se the 3 other PR's.
These are the fields where this option is used that  I know of. Maybe there are more?


GawainLynch:
Combining with #754 for merge.

PR #760, opened 2017-06-25
[Site] Update to Slugify v2

GawainLynch:
This one is just a no brainer as the only BC we'd otherwise need to worry about is the second parameter passed in call to `slugify()`, and we don't use it.

Also, Slugify 2 comes with a Silex 2 SP, so we can drop **our copy** of that file.


GawainLynch:
B.:princess:.F.L-ing

PR #762, opened 2017-06-29
Fix indentation

zomars:
Or I could be wrong?


SahAssar:
Nope, you are definitely right, thanks!

GawainLynch:
Ha! I thought I got that one too … thanks mate!

zomars:
Thanks to both of you!

PR #764, opened 2017-07-04
Twig inheritance & clean up

GawainLynch:
This is a really quick finish up from last week when I fell over … back soon, but until then I hope this helps clean up & improve the helpfulness of our Twig related docs a bit more.

@bobdenotter go for it if you want to amend anything before merge … I'll hopefully be back later in the week.


GawainLynch:
@bobdenotter Probably best we don't deploy these until I am back, as the Nginx redirects need updating on the site … but if you're comfortable with that also "go for it!" :smile_cat:

PR #766, opened 2017-07-12
[3.0] Simplify controller example

GawainLynch:
This changes the docs to extend the base controller — as intended by design — giving extension developers access to the built in helpers and making like generally easier for them.

:koala: !

### DIFF
--- a/docs/extensions/intermediate/controllers-routes.md
+++ b/docs/extensions/intermediate/controllers-routes.md
@@ -67,7 +67,7 @@ class KoalaCatcherExtension extends SimpleExtension
      * @param Application $app
      * @param Request     $request
      *
-     * @return Response     
+     * @return Response
      */
     public function callbackKoalaCatching(Application $app, Request $request)
     {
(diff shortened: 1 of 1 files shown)


docs/extensions/intermediate/controllers-routes.md
```diff
@@ -169,24 +170,16 @@ use Symfony\Component\HttpFoundation\Response;
...
+class DropBearController extends Base
 {
-    /** @var Application */
-    protected $app;
-
     /**
      * {@inheritdoc}
      */
-    public function connect(Application $app)
+    public function addRoutes(ControllerCollection $c)
```
bobdenotter: Why only `$c`? We generally use more descriptive variable names.

GawainLynch: Interface [has it that way](https://github.com/bolt/bolt/blob/release/3.2/src/Controller/Base.php#L49), so it is consistent when viewed in the IDE hover.

bobdenotter: Fair enough. :-)

PR #726, opened 2017-04-26
Restucturing Installation pages

bobdenotter:
Work in Progress to restructure the pages describing the installation method. In order to make it easier for people of different levels of experience and/or with different workflows to find the most suitable method of installation for them.

PR #769, opened 2017-07-14
Rewrite of git installation page

GawainLynch:


GawainLynch:
> I bet this also touches on what I was doing in #726. I'll need to finish that one too, before 3.3 final :-)

It will, but an easy rebase for me if you need help … I'm here for you!

PR #770, opened 2017-07-15
Adding canonical with 'default' version, and version switcher will stay on the same page, if the page exists.

bobdenotter:


GawainLynch:
Cool, we'll grab fall-out during the weekend

PR #777, opened 2017-07-21
especially this file should come via https

greggles:


GawainLynch:
The irony in the fact that was taken from their original site link.

Merging into `release/3.0`, thank you :+1:

PR #782, opened 2017-07-24
Add a note about `Permalink` vs. `Unique Alias`.

bobdenotter:
Closes https://github.com/bolt/bolt/issues/6847


GawainLynch:
I'm good with this, but one question on my mind is do we want to make the "Note" a note div?

bobdenotter:
Updated! 👍

PR #783, opened 2017-07-25
Note about omiting `--ignore-version-reqs`

bobdenotter:


docs/installation/composer-create-project/creating-a-new-project.md
```diff
@@ -241,6 +241,10 @@ This lets you fake platform packages (i.e. PHP and PHP extensions) allowing you
 to emulate your production environment, or define your target platform in the
 Composer project configuration.
 
+After making these changes, run `composer update` to get the correct versions of
```
GawainLynch: Hmm … I omitted that deliberately as that is not always the case, as we instruct above on constructing the `composer.json` file prior to first run.

GawainLynch: Meh … :shipit:

PR #793, opened 2017-08-09
API doc buttons updates

GawainLynch:
Yesterday I updated the API docs generator to cover the current sub-projects used (or about to be) in core. This enables access via the "Bolt Core Development" page.

Requires #792

… there might have been some time while waiting on instruction on how to Flexbox, so I went a bit O.C.D. on a few things while at it.


SahAssar:
:+1: CDO for the win!

PR #795, opened 2017-08-10
Site updates

GawainLynch:

PR #796, opened 2017-08-10
Remove inline styles from buttons and wrap in a container div

GawainLynch:


SahAssar:
Goes with #795, will merge both when that little niggle is fixed.

SahAssar:
:+1:

PR #799, opened 2017-08-15
Rewrite "Nginx Configuration"

GawainLynch:


docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+```nginx
+location ^~ /.well-known/acme-challenge {
+    allow                      all;
+    alias                      /var/www/.well-known/acme-challenge;
+    default_type               "application/jose+json";
+}
 ```
+
+This example is shown as a guide only. For more information see the Certbot
+[documentaiton][certbot].
```
SahAssar: documentation

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+use one of these files per-pool.
+
+**NOTE:** You must enable **one** of the `fastcgi_pass` parameters, or Nginx
+will attempt to initiate a download of the `index.php` file instead of
+executing it.
+
+
+#### `acme-challenge.conf` (optional)
+
+The optional `acme-challenge.conf` file enables aliasing Certbot challenge
```
SahAssar: Not just Certbot, any implementor of acme will use this, so this should probably say something like "file enables aliasing of acme challenge requests (like for example Let's Encrypt uses)"

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+    #fastcgi_pass               unix:/run/php-fpm/www.sock;
+
+    # If using TCP sockets uncomment the next line
+    #fastcgi_pass               127.0.0.1:9000;
+}
+```
+
+
+On larger multi-CPU hosts with several busy sites, you will no doubt want to
+use several FPM pools, with each pool defining their own socket. You can simply
```
SahAssar: Why? the worker threads are spread over the cores/CPU's anyway just like nginx worker threads are, right?

Isn't the advantage of multiple pools that you can run them as different users and therefore get security isolation?

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+    deny                      all;
+}
+
+# Block PHP files from being run in upload (files), app, theme and extension directories
+location ~* /(?:extensions|files|theme)/(.*)\.php$ {
+    deny                      all;
+}
+
+# Block access to Sqlite database files, Apache .htaccess & .htpasswd files
+location ~ /\.(?:db|htaccess|htpasswd)$ {
```
SahAssar: .htaccess and .htpasswd are handled by the hidden files directive above, right?

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+    log_not_found              off;
+    expires                    max;
+    add_header                 Pragma public;
+    add_header                 Cache-Control "public, mustrevalidate, proxy-revalidate";
+
+    try_files                  $uri $uri/ /index.php?$query_string;
+}
+
+# Don't log, and do cache, asset files
+location ~* ^.+\.(?:atom|bmp|bz2|css|doc|eot|exe|gif|gz|ico|jpe?g|jpeg|jpg|js|map|mid|midi|mp4|ogg|ogv|otf|png|ppt|rar|rtf|svg|svgz|tar|tgz|ttf|wav|woff|woff2|xls|zip)$ {
```
SahAssar: woff2 is not in nginx default mime.types file, so we probably shouldn't configure it here either (or add a note saying to add it to mime.types)

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+
+
+#### `bolt.conf`
 
+The `bolt.conf` file define location maches common to all of your Bolt sites on
+a host.
+
+```nginx
+# Redirect requests for */index.php to the same route minus the "index.php"
+if ($request_uri ~* "^(.*/)index\.php(?>/|)(.*)$") {
```
SahAssar: https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/

`rewrite ^/index.php/(.*) /$1  permanent;` should work, right?

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+location ~* ^.+\.(?:atom|bmp|bz2|css|doc|eot|exe|gif|gz|ico|jpe?g|jpeg|jpg|js|map|mid|midi|mp4|ogg|ogv|otf|png|ppt|rar|rtf|svg|svgz|tar|tgz|ttf|wav|woff|woff2|xls|zip)$ {
+    access_log                 off;
+    log_not_found              off;
+    expires                    max;
+    add_header                 Pragma public;
+    add_header                 Cache-Control "public, mustrevalidate, proxy-revalidate";
+}
+
+# Don't create logs for favicon.ico, robots.txt, or rss.jpg requests
+location = /(?:favicon.ico|robots.txt|rss.jpg) {
```
SahAssar: What's a rss.jpg?

Googling it just turns up a bunch of rss icons :P

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+}
+
+# Block access to the app, cache & vendor directories
+#
+# NOTE: If you have one or more of 'app' , 'src', 'test' and 'vendor' as
+# sub-directories of your installation root, you should uncomment
+# this location block to prevent site visitors having access to the
+# various directories that contain executable code.
+#
+# NOTE: This approach is *not* recommended for production use.
```
SahAssar: Link to the docs explaining the different installs here would be nice.

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+The `php-fpm.conf` file define the settings for the PHP FastCGI Process Manager
+used for you Bolt site(s).
+
+```nginx
+location ~ \.php$ {
+    include                    fastcgi_params;
+    fastcgi_param              SCRIPT_FILENAME $document_root$fastcgi_script_name;
+    fastcgi_param              HTTP_PROXY "";
+    fastcgi_index              index.php;
+
```
SahAssar: `fastcgi_param HTTPS $https;` would be nice here too, see #4747 and #4853

docs/installation/webserver/nginx.md
```diff
@@ -6,106 +6,263 @@ Nginx: Configuring the virtual host
...
+    # Site root
+    root                       /var/www/sites/example.com/public;
+    index                      index.php;
+
+    # Bolt specific
+    include                    global/bolt.conf;
+
+    # PHP FPM
+    include                    global/fpm.conf;
```
SahAssar: In the other places you call it `php-fpm.conf`

docs/installation/webserver/nginx.md
GawainLynch: That's a strict forward match, so it misses sub-paths with `index.php`

docs/installation/webserver/nginx.md
GawainLynch: Ha! I just had to check, RH has it forever and I built it in with the packages I maintained (guess where I got some of this from)

docs/installation/webserver/nginx.md
GawainLynch: Yeah, and RSS is dead … but people use it on sites, it's more a "helpful" addition like the LE stuff

docs/installation/webserver/nginx.md
GawainLynch: Yes, see point about "helpful" addition … if it irks too much I can just drop it

docs/installation/webserver/nginx.md
GawainLynch: This is the docs :stuck_out_tongue_winking_eye: and an inline code block.

docs/installation/webserver/nginx.md
GawainLynch: Derp

docs/installation/webserver/nginx.md
GawainLynch: Ugh, I started with ACME, but the docs link swayed me ot specifics … still not sure it is a good idea for us to even mention it now.

docs/installation/webserver/nginx.md
SahAssar: No, I just mean what the file rss.jpg is? I don't think that rss is dead, but is that like a standard path for an rss icon? I've never seen it used.

docs/installation/webserver/nginx.md
SahAssar: Yeah, but my point is that if we are going to have a comment saying "this is bad" we should have a link to what the proper solution is.

docs/installation/webserver/nginx.md
SahAssar: Well, I'd be okay with dropping it, but if we keep it it should not mention a single application that uses it, but the protocol/standard itself IMO

GawainLynch:
Ugh … 2 am rebase before :coffee: 

Screw it, someone else can take this over :disappointed:

GawainLynch:
git reflogs are more powerful than a lack of :coffee: … let's do this.

docs/installation/webserver/nginx.md
GawainLynch: OK, no. That article has context and your assertion misses that context. While I'll agree that most the time you don't want this, by the time a request as fallen though to this point you're really splitting hairs. Also if someone has a site that is trafficked to the point where this matters, they are not reading these documentation pages :wink:

docs/installation/webserver/nginx.md
GawainLynch: It is just an attempt to "be helpful" in a typical Bolt fashion. It doesn't hurt, but I am happy to drop it or modify _something_ here, but let's try to lead people toward more power via more knowledge, i.e. open to suggestions, including dropping it, but I don't think that less is more in this case but don't have an immediately better idea.

docs/installation/webserver/nginx.md
GawainLynch: See above just now about "helpful additions" … you're right that we could probably merge and better comment, but we also get emails from people often with "why are you not …" which actually boils down to blog post consumers that don't understand what they are looking at in any depth :wink: 

I guess what the two of us should put our heads together on here is generally how to best explain some of this stuff to the "average" Bolt + Nginx user … Hmm, does that make them a "LiNgPhBo"? :rofl:

docs/installation/webserver/nginx.md
GawainLynch: Still disagreeing here … How about dropping this whole section out of the example, and having a "Extra Koalas" section (think FAQ) for things like this?

docs/installation/webserver/nginx.md
```diff
@@ -1,111 +1,257 @@
...
 
-**Note:** 502 Bad Gateway Errors
 
-If you're using UNIX sockets instead of TCP ports on your PHP-FPM installation,
-you will need to change the `fastcgi_pass` parameters to match what is set in
-your PHP-FPM configuration's `listen = `, e.g.:
+##### Block access to the app, cache & vendor directories
+
+If you have one or more of `app` , `src`, `test` and `vendor` as
+sub-directories of your installation root, you should add the following
```
SahAssar: "installation root" should be "web root", right? (I know that was in the old docs too)

docs/installation/webserver/nginx.md
```diff
@@ -1,111 +1,257 @@
...
+
+
+#### `php-fpm.conf`
+
+The `php-fpm.conf` file define the settings for the PHP FastCGI Process Manager
+used for you Bolt site(s).
+
+```nginx
+location ~ [^/]\.php(/|$) {
+    try_files                     /index.php =404;
```
SahAssar: 👍 👍 👍

docs/installation/webserver/nginx.md
SahAssar: Yeah, agreed that it won't impact anyone very much, I was mainly thinking of "best practices" and that most of the time when comparing `$request_uri` in a `if` you usually want a location block (or a rewrite if that is the goal anyway) since a location block will match against request_uri anyway, right?

docs/installation/webserver/nginx.md
SahAssar: Ah, right, gotcha, that works well then :)

docs/installation/webserver/nginx.md
SahAssar: The way the PR is now works for me :) If we come up with something else we can always change it later

docs/installation/webserver/nginx.md
GawainLynch: I think "web root" is better, yeah. The use of "installation root" goes back to v1/v2 approaches.

GawainLynch: For reference, I am dropping this section altogether. We don't support flat-structure installs officially, only provide them for people that insist on using insecure & broken web hosting setups. 

If you're configuring NGINX you're not in that group. :wink:

GawainLynch:
OK, done. For the Internet archaeologists I want to leave some notes on review & changes around "If is evil". I decided in the end to drop the following block in the name of simplicity:
```nginx
if ($request_uri ~* "^(.*/)index\.php(?>/|)(.*)$") {
    return                     301 $1$2;
}
```
The review comment about the NGINX article on [If is evil](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/) misses context, specifically in this case (from the article itself):

> The only 100% safe things which may be done inside if in a location context are:
> 
> **return** ...;
> rewrite ... last;

Other examples of using "if" from the NGINX documentation:
- [Connecting NGINX to PHP FPM](https://www.nginx.com/resources/wiki/start/topics/examples/phpfcgi/#connecting-nginx-to-php-fpm)
- [FastCGI Example: Secure your upload directory!!](https://www.nginx.com/resources/wiki/start/topics/examples/fastcgiexample/#secure-your-upload-directory)
- [Mitigating DDoS Attacks with NGINX](https://www.nginx.com/blog/mitigating-ddos-attacks-with-nginx-and-nginx-plus/)

SahAssar:
Not that it matters, but the article also mentions  "It’s generally a good idea to avoid it if possible" and "There are cases where you simply cannot avoid using an if, for example, if you need to test a variable which has no equivalent directive." and comparing a URI is clearly not one one of those since we have location blocks that do that. While the "FastCGI Example: Secure your upload directory" example uses it in this way it seems to me like it is not the recommended approach.

GawainLynch:
> There are cases where you simply cannot avoid using an if

And this is one of those cases. Can you please do me a personal favour though, and if you're going to debate these things, produce empirical arguments, not just carefully chosen snippets of documentation, and opinionated rhetoric.

OK, so one of the "handicaps" of `if` is that they will *always* be checked, so abusing them can affect location resolution on a site. Side point, one of the reasons NGINX was built in the first place was performance, that is (along with security) probably Igor's largest obsession :smile:

Another "handicap" of `if` is that it can really mess with regex matches, but one part of the problem addressed here is that the regexing priority is tricky. The other is that we also need to ensure that other PHP scripts, namely `/index.php` are still executable while `/a/index.php`, `/a/b/index.php`,  `/a/b/c/index.php`,  `/a/index.php/b/`,  `/a/index.php/b/c/`, etc, all get a second match and then overridden. Location blocks have a *priority* therefore, in some/most case they are useless.

If you feel up to reading a bit of C, the code for the [`http_rewrite_module`](https://github.com/nginx/nginx/blob/branches/stable-1.12/src/http/modules/ngx_http_rewrite_module.c#L533-L866) might explain a bit more for you :wink: 

Oh, and yes Lua can be/is a better way to go, but that is relatively new in NGINX and goes well out of scope of what we are trying to achieve here (also why I agreed it was better to just drop a lot of the "helpful extras" I had included)

PR #800, opened 2017-08-20
Remove sqlite FUD

SahAssar:
Remove sqlite FUD


bobdenotter:
In the last 24 hours we've had two people in `#general` that would have been fine with SQLite, but didn't use it, because they felt "mysql was better, because the docs say so". 

So, I'm 👍 .

PR #803, opened 2017-08-22
Extend the parser to add directives

rossriley:
Fixes https://github.com/bolt/bolt/issues/6894

Update the docs to use the extend syntax when adding directives.


GawainLynch:
OK, so minor heart attack then when I was checking the URL in this page :sweat_smile: 

Short version is @Raistlfiren's PR: https://github.com/bolt/bolt/pull/5584

PR #809, opened 2017-08-30
Update common.md

lenvanessen:
Added missing common option variant and class


docs/fields/common.md
```diff
@@ -20,6 +20,8 @@ in more detail below.
 | `pattern`  | Add a custom validation pattern to validate against when saving. |
 | `readonly` | Do not allow an editor to change the value of this field. |
 | `index`    | Add a database index for this field, only add if you know what this means. |
+| `variant`  | Change the display the field. |
```
GawainLynch: `variant` is for `text`, `date`, and `datetime` only. :smile_cat:

lenvanessen:
Hala tried to 🐨  that one in there

PR #813, opened 2017-09-09
Update geolocation.md

kevinkhill:
I don't know if that was supposed to just be in brackets or a link, but I think I linked it to the right place?


kevinkhill:
I just realized this was for the wrong branch of docs. The link is correct in 3.3

GawainLynch:
Ah no worries, I'll just do some git magic :wink:

PR #814, opened 2017-09-09
Update geolocation.md

kevinkhill:
I struggled for a bit before I found out from the console error messages that the `Static Maps API` is not enough to get the geolocation field working. The `Google Maps JavaScript API` had to be enabled in the Google Cloud Platform too. I feel like it would be worth noting this to others who might encounter the same issue.

I didn't really know where to add the note, so I just put it where it is. Please move it to where you feel is appropriate. (unless it works well there) :smile:


GawainLynch:
Thanks again! We really appreciate help tidying up some of these "gotchas" :+1:

kevinkhill:
For how much time and energy you and the team has poured into something I get to use for free, it's the least I can do. :smile: I'm happy to help!

PR #824, opened 2017-09-21
Add note about internal use of double underscore prefix

SahAssar:
See discussion in bolt/bolt#7012


GawainLynch:
> Shouldn't we just include this in the docs from 3.0 and up, though?

I was just contemplating this over coffee, and I can grab the branch jump — easy as pie — when I am back on Monday.

bobdenotter:
I can do that too, no problem.. Just asking if we agree on that. :-)

GawainLynch:
Well, `__nodes` has been there since `2.something` so 3.0 is the best :dart: IMHO

SahAssar:
Yep, going from 3.0 sounds good to me... I'm not at a computer to do the rebase right now though

PR #827, opened 2017-10-03
Add link remark in Prefix/Postfix

evertalbers:
!- I have used this in 2.2.X, not sure if it still works in Bolt 3


GawainLynch:
Ninja tweaked it a bit so it wasn't so edge-case … But thank you as always! :+1:

evertalbers:
...as Markdown! Never knew that!

GawainLynch:
It is something that got added in 3.0-dev and not clearly indicated in the docs, so yay for teamwork! :smile:

PR #832, opened 2017-10-11
Change 'exists' to 'has'

jadwigo:
I'm not really sure, but it seems that the 'exists()' call does not exist.

Feel free to ignore and delete this PR if I'm wrong, but please tell me what I'm doing wrong before that.

### DIFF
--- a/docs/extensions/filesystem/working-with-files.md
+++ b/docs/extensions/filesystem/working-with-files.md
@@ -12,7 +12,7 @@ object.
 
 ```php
     /** @var bool $exists */
-    $exists = $filesystem->exists($path);
+    $exists = $filesystem->has($path);
 ```
 
 | Variable | Type | Description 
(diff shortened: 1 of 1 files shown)


GawainLynch:
It has both. `exists()` is on the handlers, `has()` is on the filesystem, so this PR is incorrect :wink: 

```php
        /** @var \Bolt\Filesystem\FilesystemInterface $fs */
        $fs = $this->app['filesystem']->getFilesystem('files');
        if ($fs->has('myfile.txt')) {
            // do something
        }
        /** @var \Bolt\Filesystem\Handler\File $file */
        $file = $fs->getFile('myfile.txt');
        if ($file->exists()) {
            // it exists, lets delete it
            $file->delete();
        }
```

jadwigo:
Then the docs might need some clarification...

PR #836, opened 2017-10-12
Use production_error_level when debug is disabled

hansgrinwis:


GawainLynch:
Thanks, @hansgrinwis :+1: 

I have moved it to the 3.2 branch as the parameter wasn't introduced until 3.2.0.

PR #839, opened 2017-10-18
Move "Extending Twig" and "Web assets" into the intermediate directory

GawainLynch:
Mainly this is so that with, for example, the bundled extension docs, we can point to one directory for "next steps" when building an extension.

I'll handle the Nginx 301s on deploy.


bobdenotter:
Looks good to me! I'd like to merge the three extension-docs-releated PRs in soon, to keep a better eye on what's changed.

GawainLynch:
> I'd like to merge the three extension-docs-releated

This is exactly why this PR exists … I was adding a final section to #837 and wanted a clean point to, um, point someone to … the problem being the "Basic" section currently (and post this would still) have information that would be potentially confusing in the context of studying how to build a bundle.

So if this is good with you, :shipit: and I will cascade the branches, and then we can merge & deploy the others too

PR #844, opened 2017-10-23
fixed wrong class signature

ToBe998:
as discussed in #6953


ToBe998:
I think, changing the branch mixed my commit with others! How to proceed?

PR #846, opened 2017-11-01
Tiny tweak to make the example for repeaters work with a wider range of fields

bobdenotter:


GawainLynch:
This is not a blocking comment, but one for discussion nonetheless. As frequently raised by @CarsonF the `dump()` function has limited use in this case, especially with newer field types … i.e. are we really sure this is what we want to recommend?

bobdenotter:
In this case - from what I could tell - `field.value` is either a string, an array (for images, files) or a DateTime.. So, `dump(field)` gives a lot of cruft, but `dump(field.value)` gives sensible results.

GawainLynch:
> In this case - from what I could tell - field.value is either a string, an array (for images, files) or a DateTime.. So, dump(field) gives a lot of cruft, but dump(field.value) gives sensible results.

Until it is a "lazy" type, then we get things that confuse the hell out of people. But further, my concern is the section is called _"Example usage in templates"_ and that is debugging, not "example use"

bobdenotter:
Updated. :-)

GawainLynch:
Perfect … I think that's a lot clearer on intent :+1:

PR #847, opened 2017-11-01
Better order for field types list.

bobdenotter:


GawainLynch:
Another "not a blocker" question … Sanity checking the ordering here, I can't see the logic.

bobdenotter:
I felt the ordering was quirky, so i tried re-ordering it to what I thought makes sense, as to how regularly the fields are used, from what I hear / see in other people's projects.

GawainLynch:
WFM

PR #848, opened 2017-11-01
Initial docs for `block` fields.

bobdenotter:


GawainLynch:
This is branched on top of #847 so I'll hold off on this pending the discussion there.

PR #852, opened 2017-11-09
add template select to list of fields

t14:
I know this page exist under the template section. But it is not obvious.

Thinking contextually If I am looking for documentation on templateselect field type. I would go to the field types page and not the templateing section.

### DIFF
--- a/docs/fields/index.md
+++ b/docs/fields/index.md
@@ -21,6 +21,7 @@ pages:
     - imagelist
     - filelist
     - geolocation
+    - templateselect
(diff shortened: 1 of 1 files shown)


bobdenotter:
This will need slightly more than just that link added.. 

I'll see if i can whip up a new page soon. Thanks for pointing it out, @t14 ! 👍

bobdenotter:
Turns out the page indeed is present already, just not linked.. PR is GTG.

PR #860, opened 2017-11-21
Fix internal link to thumbnail

evertalbers:
Maybe there's other places in the docs that link to this?


GawainLynch:
A quick check says this is the only one. Thanks, @evertalbers :+1:

PR #865, opened 2017-11-27
Added further instructions for nginx configurations.

itsa-sh:


GawainLynch:
Thanks, @itsash-- :+1: 

Wondering if we could expand this slightly (happy to do it, but you're there so…) and indicate subtly why "20 MiB" is the chosen value?

PR #869, opened 2017-12-01
Migrate site to Symfony 4

GawainLynch:
Let's test this a bit first, and discuss approach. :smile:


src/Controller/Site.php
```diff
@@ -132,59 +103,43 @@ public function sitemap(Version $version)
...
      */
-    public function error(\Exception $e, Request $request, $code)
+    public function error(\Exception $e, Request $request, Documentation $documentation, $code)
     {
-        $requestUri = explode('/', $request->getRequestUri());
+        $requestUri = \explode('/', $request->getRequestUri());
 
         // Don't trap Symfony shizzle.
-        if (in_array($requestUri[1], ['a', '_profiler']) || $this->app['debug']) {
+        if (\in_array($requestUri[1], ['a', '_profiler'], true) || $this->app['debug']) {
```
CarsonF: Missed `$this->app['debug']`

src/Controller/Site.php
```diff
@@ -1,66 +1,36 @@
...
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
-class Controllers implements ControllerProviderInterface
+/**
+ * Site controller.
+ *
+ * @author Carson Full <carsonfull@gmail.com>
+ * @author Gawain Lynch <gawain.lynch@gmail.com>
+ */
+class Site extends AbstractController
```
CarsonF: Shouldn't `Documentation` be passed into the class instead since it is needed for every method?

src/ArgumentResolver/VersionValueResolver.php
```diff
@@ -0,0 +1,50 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Controller\ArgumentValueResolverInterface;
+use Symfony\Component\HttpKernel\ControllerMetadata\ArgumentMetadata;
+
+/**
+ * Version value resolver.
+ *
+ * @author Gawain Lynch <gawain.lynch@gmail.com>
+ */
+final class VersionValueResolver implements ArgumentValueResolverInterface
```
CarsonF: This is a lot of code just to skip a getter...WDYT?

A `Documentation::getVersionOrDefault` could remove the conditional logic in each place.

src/Documentation.php
```diff
@@ -38,31 +47,32 @@ public function __construct(Yaml\Parser $yamlParser, PageBuilder $pageBuilder, $
...
     }
 
-    public function load()
+    public function load(): void
     {
         $dirs = (new Finder())->directories()->in($this->versionDir)->sortByName()->depth(0);
 
         foreach ($dirs as $dir) {
-            /** @var SplFileInfo $dir */
+            /* @var SplFileInfo $dir */
```
CarsonF: Derp

src/PageBuilder.php
```diff
@@ -129,28 +133,28 @@ protected function loadCollection($dir)
...
                 $page = $this->loadPage($file);
-                $str = serialize($page);
+                $str = \serialize($page);
                 $cache->write($str, [new FileResource($this->root . $file)]);
             }
         );
 
         if (!$page) {
-            $page = unserialize(file_get_contents($cache->getPath()));
+            $page = \unserialize(\file_get_contents($cache->getPath()));
```
CarsonF: Want to change these `serialize` and `unserialize` calls to use `Bolt\Common\Serialization`?

src/Version.php
```diff
@@ -25,37 +33,30 @@ class Version
...
+     * @param PageBuilder $builder
+     * @param Yaml\Parser $yamlParser
      */
-    public function __construct(
-        $version,
-        $basePath,
-        PageBuilder $builder,
-        Yaml\Parser $yamlParser
-    ) {
+    public function __construct($version, $basePath, PageBuilder $builder, Yaml\Parser $yamlParser)
```
CarsonF: Typehints?

CarsonF:
Also, what about using PHP DI instead of YAML?

src/Documentation.php
GawainLynch: That's something coming from the CS fixer … NFI what to do about it yet, meant to bring it up with you in Slack … but away from home ATM.

src/PageBuilder.php
GawainLynch: Derp … totally missed these :disappointed:

src/Version.php
GawainLynch: Done, plus the one you missed ;-)

src/Controller/Site.php
GawainLynch: Yeah, knew about that one … was coming back for it this morning :+1:

GawainLynch: … had to implement the 404 handling still

src/Controller/Site.php
GawainLynch: WFM … done

src/ArgumentResolver/VersionValueResolver.php
GawainLynch: 13 LLoC … :wink: 

But I'll cover the reasoning below

GawainLynch:
> Also, what about using PHP DI instead of YAML?

That was my first desire too … But as with the value resolver, I want to try to get the team up-to-speed with what can be done without getting too complex. The stuff I am doing in the core prototype is going to arrive in PHP, but this was some low hanging fruit I wanted to put on the table now so that people can start to get comfortable with the differences between Pimple and SF4 without too much of a curve.

GawainLynch:
> I first saw the lines changed stat and thought WTH...but it's just the lock files lol. I wish there was a good way to exclude those from stats.

Oh wow, I didn't quite grasp what you meant until I just refreshed the page and noticed those numbers myself … :scream_cat: … but yeah, the PHP changes are pretty tiny for the most part and it the weight of the delta is in JavaScrap moved & lock files.

bobdenotter:
I'm not too fond of the `webpack.config.js` and `yarn.lock` being in the root. Can we move all asset-build-node-stuff into a separate folder.. Not sure what the best name would be, though.

GawainLynch:
> I'm not too fond of the webpack.config.js and yarn.lock being in the root

Welcome to Flex :wink:

bobdenotter:
> Welcome to Flex 😉

As in "deal with it", or "That's just the default/convention"?

It seems like it wouldn't be hard to change this, for us.

GawainLynch:
>> Welcome to Flex :wink:

> As in "deal with it", or "That's just the default/convention"?

Oh sorry, that might have been a bit terse … OK, so what Flex does is "try" to manage things for you, and for those two it assigns them to the project's root directory … yes, they can be moved but that un-Flexes them.

As it stands presently, Flex is a little more opinionated than some are going to like. I think this will change over time, as there was a short time to deliver a complex solution, i.e. Flex.

Personally, I am going with the flow on how things are at the moment with the view to adapt as Flex does … my thinking is that I really don't want to code NIH into anything if I/we can help it … we don't know what/how this is going to develop, etc.

Not pushing the above point, just laying out my thinking and not wishing to invest large amounts of time into people's ideology … "you" want to change it, fine, but "you" immediately take ownership of it :wink:

bobdenotter:
Fair enough. If this location makes it easier to keep in sync, i've no objections. 

Let's leave it like this for now. Once the updated frontend work is about ready to be merged back in, we can always revisit it we want to keep it as is, or move it around. :-)

GawainLynch:
Yeah, that is pretty much where my thinking it at … "perfect" is great and all, but let's lower the maintenance barrier and be prepared that things are going to change (probably in the direction we both desire) pretty soon with Flex, etc. One thing I was expecting was that Symfony would hit a similar thing that we do … ask people to kick the tires and test pre-major, but it isn't until release you end up with a dot-zero :rofl: 

FWIW the location of those files in the root sends my O.C.D. off the charts, but I am balancing that against how many people are willing and able to help with this versus hour in the day :wink:

src/Documentation.php
CarsonF: 🤦‍♂️ Yeah I remember that now. It might be a comment fixer to tweak

src/ArgumentResolver/VersionValueResolver.php
CarsonF: > also, this is essentially what was being done previously with the ->convert()

Yeah I know. My initial hope when coding was that `Documentation` class wouldn't be needed in the controller methods and the selected `Version` would be enough. But it turned out not to be the case.

bobdenotter:
Bumping this to SF 4.0.1, I noticed a new file, `config/routes/annotations.yaml`, and a change in `.gitignore`. 

`@symfony/webpack-encore` was still on `0.12`. I bumped this to `0.17`, and tried to build the assets. Is it OK if i push some fixes (like a missing path to Foundation) to this branch?

GawainLynch:
> Bumping this to SF 4.0.1, I noticed a new file, config/routes/annotations.yaml, and a change in .gitignore.

That was already rebased in … you'll need to reset against this branch … those are now caught and cleaned up, along with about a dozen bugs I found this morning.

> and tried to build the assets

Yep, left that in "known broken" as I didn't want to invest time on that part … see below …

> Is it OK if i push some fixes (like a missing path to Foundation) to this branch?

But of course! As discussed on Slack, the UI stuff I am leaving to others, I'm just getting this moving.

PR #878, opened 2018-01-24
Update task scheduler docs to show timed record usage

rossriley:


GawainLynch:
Added in https://github.com/bolt/bolt/pull/5680

PR #879, opened 2018-01-30
New documentation theme

dvangimst:
Added new documentation theme according to the same design as extension docs.

bolt docs - new design


GawainLynch:
Thanks very much for taking this on, @dvangimst :+1: 

OK, a couple of things (noting the `WiP` in the title):

- Assets sources should be in `assets/` and not `source/`
- Gulp's gotta go, and Webpack restored
- `package-lock.json` should go and be replaced with a `yarn.lock` — "Say no to NPM!"
- There should be no `vendor/` they can be added as dependencies, and compiled in as modules
- Fontawesome 5's `all.js` file is ~640 KiB for a few icons … I'm not at all in favour of that approach for a documentation site, especially one that gets as much mobile use as this.

I can help with all of these, just call out on Slack … happy to add commits or advice … your call :smile:

GawainLynch:
P.S. I'm more than slightly excited to see this getting close to the light-of-day :tada:

GawainLynch:
Follow-up testing:

- Request count drops from 21 to 20 but Webpack can reduce that to 16 requests (still a lot)
- Transfer load goes from 741 KiB to just over 1.5 MiB
- Un-cached complete front page loads go from < 1s to ~3s (too many requests, and larger downloads that won't reliably cache)

Probably out-of-scope, but the jQuery JS is 40% of the JS download total and only needed for Magnific, which itself is not useful. Can we use something slimmer without jQuery?

GawainLynch:
Stipping jQuery out, using vendors as vendors,  and Webpacked:

- 14 requests
- 602 Kib
- 0.72s
- 20 minutes work

Friends don't let friends use jQuery.

GawainLynch:
@dvangimst I  have pushed a squashed update, there are still a couple of things b0rken in the SASS, but @bobdenotter was happy to hand over at that point.

Still to do is porting the JS in `legacy.js` to native JS, and what to do with replacing (?) Magnific.

GawainLynch:
Oh, and see the `assets/README.md` file for super-simple build instructions.

Also, note that I have left out compiled CSS/JS as scores of un-rebase-able asset commits will only make this hard on everyone :smile:

bobdenotter:
Todo for now: 

- [ ] Make Prismjs and "Copy to clipboard" work.
- [ ] Make Algolia search work.
- [ ] Make 'version changer' work
- [ ] Make 'slideToggle' work.
- [ ] Replace Magnific Popup with something.
- [ ] Add 'stack' class to tables, for mobile
- [ ] Make "Hamburger" menu work on mobile

rossriley:
👍

PR #880, opened 2018-01-31
Add prefix/postfix examples

evertalbers:
I think this is enough to make people realise that these options exist.


GawainLynch:
Thanks mate :+1:

PR #881, opened 2018-02-09
Use setZone() function to define the Widget Zone

nestordedios:
Using setType() function is not possible. That function does not exist in [Widget.php](https://github.com/bolt/bolt/blob/3.4/src/Asset/Widget/Widget.php) but correct me if I'm wrong. 

I think this changes can be merged into 3.0 and up.


rossriley:
🐨

PR #883, opened 2018-02-18
Fix the color of "Search the documentation"

SahAssar:
This is pretty untested, I don't have access to a proper system ATM, but the contrast levels between the label and the background were really annoying me :)

### DIFF
--- a/assets/scss/modules/_search.scss
+++ b/assets/scss/modules/_search.scss
@@ -2,6 +2,10 @@
 
 $search-height: 55px;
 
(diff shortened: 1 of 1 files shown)


GawainLynch:
There are colours?

SahAssar:
@GawainLynch yeah, some bastard went and invented them in 1861: https://en.wikipedia.org/wiki/Color_photography

GawainLynch:
Note to self: prior to 1861 we saw everything in B&W :+1:

PR #885, opened 2018-02-23
Fix config keys, add an example

panda-madness:
Fix incorrect keys for the session configuration in `config.yml/config_local.yml`. Also provides a working example.

Bashed my head on this for a while before looking at the source code.


panda-madness:
Added a Memcached example, would be nice if you could verify it's correctness.

PR #893, opened 2018-04-06
Added a useful ckeditor option

nealmarques:
I thought people might find it useful to be able to remove some of the buttons from ckeditor.  I listed all the bolt default ones.

### DIFF
--- a/docs/fields/html.md
+++ b/docs/fields/html.md
@@ -45,6 +45,7 @@ field.
                     autoGrow_onStartup: false
                     uiColor: '#BADA55'
                     autoGrow_minHeight: 300
+                    removeButtons : "Bold,Link,Unlink,NumberedList,BulletedList,Italic,Format,Indent,Outdent,RemoveFormat,Maximize" # This would remove all the default bolt buttons from the button bar.
 ```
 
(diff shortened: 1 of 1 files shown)


SahAssar:
If possible I'd like for this to use the "folded style" or "literal style" of yaml (see when CTRL-f "fold" here: http://symfony.com/doc/current/components/yaml/yaml_format.html) so that we don't get the excessive line-length and can easier list the options. It'd also allow for nice git diffs 😄

nealmarques:
Thing is though, white space is actually important inside those quotes.  You add a space before after any comma there, and it the button will appear... if I'm even following what you are saying.

SahAssar:
Right, that's what I meant by "if possible". If it's not then don't, I just wanted to check if it is 😄 Perhaps we can have the comment on the preceding line though? Something like

```
# The next line controls which of the bolt default formatting buttons should be shown
```

bobdenotter:
Merged into 3.0, in 1660d3cb183dc32f8b87d7a3a62a75ab4c34b76c. 

Thanks, @nealmarques 

I'll ninja-edit the comment, as @SahAssar suggested. 👍

PR #896, opened 2018-04-18
Add information about cache folder outside of site root

jadwigo:
Adds an explanation in docs for https://github.com/bolt/bolt/issues/7440

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -9,4 +9,5 @@ pages:
     - thumbnails
     - routing
     - advanced
+    - troubleshooting
 ---
--- a/docs/configuration/troubleshooting.md
+++ b/docs/configuration/troubleshooting.md
@@ -1,0 +1,60 @@
+---
+title: Troubleshooting configuration
+---
+Troubleshooting configuration
+=============================
+
+Cache folder outside of the app folder
+--------------------------------------
+
+Normally the `cache` folder is inside of the `app` folder and the site
+structure will look like this:
+
+```
+.
+└── example.org/
+    ├── app/
+    │   ├── cache/
+    │   ├── config/
+    │   └── database/
+    ├── extensions/
+    ├── public/
+    └── vendor/
+```
+
+In some cases you want the cache folder to be outside of that structure, for
+example if you have a deploy process that has a shared folder. In that case you
+want to use an absolute path for the cache folder.
+
+
+```
+.
+├── current -> releases/2
+├── releases/
+│   ├── 1/
+│   │   ├── app/
+│   │   ├── extensions/
+│   │   ├── public/
+│   │   └── vendor/
+│   └── 2/
+│       ├── app/
+│       ├── extensions/
+│       ├── public/
+│       └── vendor/
+└── shared/
+    ├── cache/
+    └── files/
+```
+
+The Bolt composer update or installation will attempt to update the paths in
+`.bolt.yml`. If the cache path is absolute it will add an `app` path that will
+look like `app: %site%/../../shared/cache` or similar. You have to manually set
+the `app` path in the `.bolt.yml` configuration file to `app: %site%/app` for
+Bolt to work.
+
+```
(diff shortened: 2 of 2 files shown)


docs/howto/troubleshooting-outside-webroot.md
```diff
@@ -261,6 +261,60 @@ line with the `require` in it, and change it like this:
...
+│       ├── app/
+│       ├── extensions/
+│       ├── public/
+│       └── vendor/
+└── shared/
+    ├── cache/
+    └── files/
+```
+
+The bolt composer update or installation will attempt to update the paths in `.bolt.yml`. if the
```
xiaohutai: `bolt` -> `Bolt`

docs/howto/troubleshooting-outside-webroot.md
```diff
@@ -261,6 +261,60 @@ line with the `require` in it, and change it like this:
...
+cache path is absolute it will add an `app` path that will look like `app: %site%/../../shared/cache`
+or something similar. You have to manually set the `app` path in the `.bolt.yml` configuration file
+to `app: %site%/app` for bolt to work.
```
xiaohutai: `bolt` -> `Bolt`

jadwigo:
The behaviour did occur on updates from 3.3.x to 3.4.10+ or updates from 3.3.x to 3.5.x so I assumed it was only relevant for bolt 3.4 / 3.5

As for the location in the docs for troubleshooting outside the webroot - that was a bit of a stretch. It was the closest match for where I would look. There is no "miscellaneous troubleshooting" or "advanced configuration with bolt.yml" document yet, so I put it here.

GawainLynch:
> The behaviour did occur on updates from 3.3.x to 3.4.10+ or updates from 3.3.x to 3.5.x so I assumed it was only relevant for bolt 3.4 / 3.5

Yeah, I've been working around it since late 3.0.0-beta on my sites, and I mentioned that to @bobdenotter in Slack but it didn't make it to you.

> As for the location in the docs for troubleshooting outside the webroot - that was a bit of a stretch. It was the closest match for where I would look. There is no "miscellaneous troubleshooting" or "advanced configuration with bolt.yml" document yet, so I put it here.

Yeah, I have to think through it myself. Thing is that it doesn't just affect cache, but also the database directory is another one that will catch you out if you're using Sqlite.

jadwigo:
How about moving the new section to `configuration/troubleshooting-configuration.md`

GawainLynch:
I was thinking `docs/installation/composer-create-project/creating-a-new-project.md` as that is where `.bolt.yml` is documented

jadwigo:
For me `.bolt.yml` is used in more than just projects created with composer - I need it in sites managed with deployer too.

GawainLynch:
That is an implementation detail for you though.

jadwigo:
It does not mean that I'm set on keeping the page where I put it... feel free to move and modify as needed.

JarJak:
Docs in this PR are too exceptional. It should cover more cases. Example:
Before:
```
paths:
    cache: %site%/var/cache
    database: %site%/var/database
```
After:
```
paths:
    app: '%site%/%site%/var'
```

So this is wrong:
>If the cache path is absolute it will add an `app` path that will look like `app: %site%/../../shared/cache` or similar.
1. It will not only add `app` but also remove other paths from `.bolt.yml`
2. This happens not only with absolute paths.

jadwigo:
Oh sorry... I deleted the branch, not realizing it was still in use.

JarJak:
@gawainlynch can you reopen this?

GawainLynch:
Unfortunately no, as the branch is no longer so GitHub greys out the reopen button. But also the details were a long way off anyway, and there would be less effort just using the diff for inspiration and starting fresh :innocent:

jadwigo:
@JarJak if you have the time, please make a new pull request for the docs, I'm unable to at the moment

PR #902, opened 2018-05-09
Solving: Apache not reading .htaccess

ColonelBlimp:
Added some pointers to solving Apache not reading the .htaccess file.


GawainLynch:
Thanks, @ColonelBlimp

PR #906, opened 2018-05-18
Performance: Mention autoloader optimization

jnvsor:
I've seen production websites more than halve their response time
by optimizing the autoloader, and the base install of bolt 3.5
gives a 20% speed boost after optimizing on my development machine.


bobdenotter:
Thanks, @jnvsor 👍

PR #920, opened 2018-08-14
Removed 2nd "templateselect"

hansgrinwis:
Templateselect appeared twice in the fields index, causing misbehaving pagination. Removed first instance.

### DIFF
--- a/docs/fields/index.md
+++ b/docs/fields/index.md
@@ -12,7 +12,6 @@ pages:
     - markdown
     - textarea
     - repeater
-    - templateselect
     - block
     - integer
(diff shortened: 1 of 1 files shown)


bobdenotter:
Good catch.. It seems to appear here first: https://raw.githubusercontent.com/bolt/docs/3.4/docs/fields/index.md

I'll see how I can best fix that. :-)

bobdenotter:
@hansgrinwis I've merged the change into `3.4`, and then merged the changes down. It'll be included in the next update on the site. Thanks!

PR #923, opened 2018-08-18
Twig docs links from  Twig 1.x to Twig 2.x

hansgrinwis:
Example:
https://twig.symfony.com/doc/1.x/functions/index.html
changed to
https://twig.sensiolabs.org/doc/functions/index.html 
which will (currently) redirect to 
https://twig.symfony.com/doc/2.x/functions/index.html

### DIFF
--- a/docs/templating/building-templates.md
+++ b/docs/templating/building-templates.md
@@ -299,13 +299,13 @@ What happens in this example is the following:
 
 [twig]: http://twig.sensiolabs.org/documentation
 [twig-designers]: http://twig.sensiolabs.org/doc/templates.html
-[control-structure]: https://twig.sensiolabs.org/doc/1.x/templates.html#control-structure
-[twig-tags]: https://twig.sensiolabs.org/doc/1.x/tags/index.html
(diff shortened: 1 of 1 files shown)


bobdenotter:
Bolt uses Twig 1.x, so i don't think we should change these. :-)

hansgrinwis:
Ow? Ok! Other links on that page link to the Twig docs using URLs w/o the version number, so those will direct to the 2.x docs. Shall I modify these (back) to 1.x?

bobdenotter:
Well, changing them back isn't needed, right? Or did you want to change them to something else than . they are in the code, now?

hansgrinwis:
No. What I mean is: there are two other links to the Twig docs, that I did not modify:

[twig]: http://twig.sensiolabs.org/documentation
[twig-designers]: http://twig.sensiolabs.org/doc/templates.html

Shall I modify these (back) to 1.x?

bobdenotter:
> Shall I modify these (back) to 1.x?

In that case: Yes, please. :-)

PR #924, opened 2018-08-18
Theme.yml info corrected

hansgrinwis:
Deleted 'template selection' part, as it has a separate page. The info here was not totally correct anyway.
Changed 'theme.yml' because theme.yml has priority over 'config.yml'.

### DIFF
--- a/docs/templating/building-templates.md
+++ b/docs/templating/building-templates.md
@@ -171,6 +171,18 @@ helper templates by default, when providing you with a list to select your templ
 <code>include</code> tag to insert the header, footer and such, but you're free to use
 <a href="template-inheritance">Template Inheritance</a> if you prefer.</p>
 
+By default, Bolt creates links to single pages based on the ContentTypes, and
+it uses a template based on its name.
+
+For instance, if your site has a ContentType `foos`, a single record in that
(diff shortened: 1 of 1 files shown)


bobdenotter:
The typos and removed duplicates look good, but why did you remove the section about `foo.twig`. That still works, right?

hansgrinwis:
I deleted that because it is not complete info. The complete info is available in the page before this one (the one on template selection).

However, you made me thinking: if you start with no template defined anywhere, Bolt will try to find `foo.twig`, and if `foo.twig` does exist, the only way to override that is via the Contenttype (either by `listing_template` or a `templateselect` field). It confused me that the template selection from (`theme.yml` or) `config.yml` was not mentioned here, but that only applies once`foo.twig` does not exist.

I will undelete that part, when I have access to a decent workstation. (Currently typing on my phone.)

Thanks.

bobdenotter:
Ok, sounds good. When you've updated the PR, i'll take another look! 👍

hansgrinwis:
This is better, I think...

bobdenotter (approved):
Looks good to me now! Thanks, @hansgrinwis 👍

PR #925, opened 2018-08-19
Template selection from theme.yml

hansgrinwis:
Templates defined in `theme.yml` override those defined in `config.yml`.


bobdenotter (approved):
👍

PR #930, opened 2018-08-27
Updating fontawesome references

Boorj:
**Font Awesome** updated their assets and website. So it would be nice to keep links updated (and consistent).


bobdenotter (approved):
👍

PR #939, opened 2018-10-05
Update usage of image filters.

artggd:
Hi,

Most of my Bolts don't have any alts on `img` tags. Today I learned this is because the image filters must be used with `record.value.cover` not just `record.cover`.

Updating this part of the docs would probably save some time to a few.


bobdenotter (approved):
Thanks!

PR #940, opened 2018-10-13
Mention `separator` in docs

### LINKED ISSUE #935: field: separator:true  should be mentioned
There's an option `separator: true` which creates a Divider. It's being set up in field in contenttypes.yml. It's not documented, but it exists. Here's the code from `_field.twig`:

```twig

{# Postfix #}
{% if field.postfix is defined and field.postfix is not empty %}
    <div class="postfix">
        {{ field.postfix|markdown }}
    </div>
{% endif %}

{# Divider #}
{% if field.separator is defined and field.separator == true %}
    <hr>
{% endif %}
```
In docs there's even an advice to add <hr> manually.
> As you can see, using `postfix: "<hr>"` gives a simple and effective way of adding a divider in the edit screen.

So i think separator deserves to be documented. That could be fixed in docs, in _common_ and _nearby 'Postfix and prefix'_ section. And in cheatsheet too.

#### Places to edit:
**In docs:**
https://github.com/bolt/docs/blob/3.0/docs/fields/common.md#common-options-for-all-fields
and
https://github.com/bolt/docs/blob/3.0/docs/fields/common.md#prefix-and-postfix

**In cheatsheet:**
https://github.com/bolt/docs/blob/3.0/cheatsheet.yml#L135

Boorj:
closes #935

### DIFF
--- a/cheatsheet.yml
+++ b/cheatsheet.yml
@@ -134,6 +134,7 @@ fieldoptions:
         - [ "`group: Foo`", "Separate this and all subsequent fields (until another group declaration) into a tab with the name 'Foo'. Replace 'Foo' with the desired label of the tab." ]
         - [ "`prefix: ..`", "HTML to add before the field. " ]
         - [ "`postfix: ..`", "HTML to add after the field. " ]
+        - [ "`separator: true`", "Add `<hr>` divider after the field. " ]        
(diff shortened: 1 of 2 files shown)


bobdenotter (approved):
Thanks!

PR #943, opened 2018-10-31
adds documentation on using json_manifest with the asset twig function

cdowdy:
for https://github.com/bolt/bolt/issues/7483


cdowdy:
made pull against wrong branch.. fixing :)

PR #948, opened 2019-02-15
Documentation for custom image attributes (Bolt PR 7743)

ankedsgn:
Adds documentation for the custom attributes for imagefields.
Goes with: https://github.com/bolt/bolt/pull/7743

### DIFF
--- a/docs/fields/image.md
+++ b/docs/fields/image.md
@@ -35,17 +35,18 @@ field.
 
 * `extensions` Allows you to restrict users to only be able to upload files with
   certain file extensions.
-* `attrib` Can be set to a custom attribute, such as alt-texts, captions or data-attributes. For example:
+* `attrib` Can be set to a custom attribute, such as alt-texts, titles, captions or data-attributes. For example:
   * `alt` Show a text field for the `alt` parameter
-  * `caption` Show a text field for the `caption` parameter
-  * `[alt, caption]` Add more then one attribute
+  * `title` Show a text field for the `title` parameter
+  * `caption` Show a custom field for a `caption` this image uses in your template
+  * `[alt, title, caption]` Add more then one attribute
 * `upload` Allows you to upload files for this field into a specified directory
   so they remain grouped. This directory will be created in `{%web%}/files/`
 
 ```yaml
         cover:
             type: image
-            attrib: [alt, caption]
+            attrib: [alt, title, caption]
             extensions: [ gif, jpg, png ]
             upload: portfolio
 ```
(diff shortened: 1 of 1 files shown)


docs/fields/image.md
```diff
@@ -35,21 +35,21 @@ field.
 
 * `extensions` Allows you to restrict users to only be able to upload files with
   certain file extensions.
-* `attrib` Can be set to either of the following:
+* `attrib` Can be set to a custom attribute, such as alt-texts, captions or data-attributes. For example:
   * `alt` Show a text field for the `alt` parameter
-  * `title` Show a text field for the `title` parameter
```
JarJak: title is still a special attribute so I would leave it here

JarJak (changes requested):

ankedsgn:
Title is discouraged to be used due to accessibility reasons, so I opted to go for the more general 'caption' instead, thus showing a practical example.

JarJak:
Let's leave title (since it is special in Bolt's code) together with alt and caption. You can also add a note that title is discouraged if you want.

ankedsgn:
> Let's leave title (since it is special in Bolt's code) together with alt and caption. You can also add a note that title is discouraged if you want.

I agree, It's handled specifically in Bolt's code so it kinda deserves a mention. I put this back in the docs. Adding a note on a11y might be a bit much, i am not sure if it's the right place here. :)

docs/fields/image.md
```diff
@@ -35,21 +35,22 @@ field.
 
 * `extensions` Allows you to restrict users to only be able to upload files with
   certain file extensions.
-* `attrib` Can be set to either of the following:
+* `attrib` Can be set to a custom attribute, such as alt-texts, titles, captions or data-attributes. For example:
   * `alt` Show a text field for the `alt` parameter
   * `title` Show a text field for the `title` parameter
-  * `[title, alt]` Show both fields
+  * `caption` Show a custom field for a `caption` this image uses in your template
+  * `[alt, title, caption]` Add more then one attribute
```
JarJak: than*

docs/fields/image.md
```diff
@@ -35,21 +35,22 @@ field.
 
 * `extensions` Allows you to restrict users to only be able to upload files with
   certain file extensions.
-* `attrib` Can be set to either of the following:
+* `attrib` Can be set to a custom attribute, such as alt-texts, titles, captions or data-attributes. For example:
   * `alt` Show a text field for the `alt` parameter
   * `title` Show a text field for the `title` parameter
-  * `[title, alt]` Show both fields
+  * `caption` Show a custom field for a `caption` this image uses in your template
```
JarJak: that image will use in your template*

JarJak (changes requested):

docs/fields/image.md
ankedsgn: meh. f*ck. :)

JarJak (approved):

JarJak:
Thank you @anketwokings !

ankedsgn:
@JarJak Thanks for taking the time to check both the PRs! Have a good weekend!

PR #951, opened 2019-03-03
Updated documentation for .bolt.yml regarding view

cscharfen:
If view is set to bolt-public/view, the layout does not work at all. Removing /view from this entry, solves the issue. (Which makes sense, because there is no "view" directory in bolt-public.)
It is also mentioned in this answer: https://discuss.bolt.cm/d/298-changing-folder-paths-for-database-config-etc/4

### DIFF
--- a/docs/howto/troubleshooting-outside-webroot.md
+++ b/docs/howto/troubleshooting-outside-webroot.md
@@ -168,7 +168,7 @@ paths:
     web: www
     themebase: www/theme
     files: www/files
-    view: www/bolt-public/view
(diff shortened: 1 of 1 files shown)


cscharfen:
I just realized, that I did the change for the 3.6 branch, but it should be merged into the master branch too.

JarJak:
@cscharfen 3.6 is the actual "master" branch

JarJak (approved):

bobdenotter (commented):
Thanks! 👍

PR #944, opened 2018-11-06
adds documentation on using json manifest with the asset twig function

cdowdy:
for https://github.com/bolt/bolt/issues/7483  

since this was added to bolt 3.5 this isn't available for all of 3.x series unfortunately 

relevant pull request where this was added https://github.com/bolt/bolt/pull/7305

PR #965, opened 2019-10-08
add missing doc about using json manifest with asset function

dadaxr:
that doc was previously created and merged in https://github.com/bolt/docs/pull/944
but was never merged into 3.X and 3.6 branch...


bobdenotter (approved):
Good catch! Thank you! 👍

PR #972, opened 2019-10-25
Rewriting the documentation to easily step-by-step (cue NKOTB) ...

ankedsgn:
... and scannable docs.

The 4.0 docs are now still mainly (wall of) text based, which is good as a start. But to make it easy for reference we want to add more headers, steps, examples, short sentences and whitespace. This is a work in progress.


bobdenotter (approved):
Looking good so far! 👍

PR #987, opened 2020-01-31
Link to composer install page

evertalbers:
(presuming that the link should be changed and not the filename of the target)


bobdenotter (commented):
Good catch, thanks!

bobdenotter (approved):

PR #989, opened 2020-02-03
Images contained in updated v4 docs

I-Valchev:


bobdenotter:
If we delete existing files, won't that break if you back to older versions of the documentation?

I-Valchev:
hmmm yes that's a very good point. Also if we replace images, I assume it'll show the replaced files in older versions, which shouldn't really happen. Let me work on this a little more :-)

PR #993, opened 2020-02-03
change to apply spaceless

0spinboson:


bobdenotter (approved):
Thanks, good catch! 👍

PR #1017, opened 2020-03-13
Use the .htaccess code from the distribution file in the docs

evertalbers:
I suggest this because to make the  "block from lines 22-35" (near the end of this markdown file) mention the correct line numbers again. Not sure if this is what you intended though.


bobdenotter (approved):
Looks good! Thanks, @evertalbers

PR #1023, opened 2020-03-29
Fix routing examples

technicallyerik:
Update the routing examples to make sure they all work with the latest version of Bolt 4.


bobdenotter (approved):
Awesome, thank you! 👍

PR #1026, opened 2020-03-31
Delete set.md

eduardomart:
This is no longer needed since Collections are the new name

### DIFF
--- a/docs/fields/set.md
+++ b/docs/fields/set.md
@@ -1,3 +1,4 @@
+# DELETE
 ---
 title: Set field
 ---


bobdenotter:
Hmm, not really. You can still use Set to group together a few fields. In fact, a Collection is a group of Sets. While we foresee them used most often in Collections, a singular Set is still a thing people can use.

Maybe we should clarify this better in the docs, though? 🤔 💭

eduardomart:
Yeah, I am confused now. 
So are Sets =  Repeaters(V3) and  Collections = Blocks(v3) ?
Wasn't the idea of Collections to just have one field type since Repeaters and Blocks were basically the same thing?

bobdenotter:
I made a small diagram.. Does this clarify it? 

```
                                         ┌─────────────┐    ┌─────────────┐                               
                                         │    Title    │    │   Heading   │                               
                                         ├─────────────┤    ├─────────────┤                               
     Let's say we have two groups of     │             │    │             │                               
    fields that go together, and are     │    Image    │    │    HTML     │                               
             used in unison:             │             │    │             │                               
                                         ├─────────────┤    ├─────────────┤                               
                                         │  Link text  │    │   Button    │                               
                                         └─────────────┘    └─────────────┘                               
                                                                                                          
                                                                                                          
                                                                                                          
                                                                                                          
                                          ┌───────────────────┐                                           
                                          │      Repeater     │                                           
                                          │                   │    We can repeat this                     
                                          │ ┌─────────────┐   │  *one* group of fields,                   
 This is how it works in Bolt 3:          │ │    Title    │   │    one or more times                      
                                          │ ├─────────────┤   │                                           
                                          │ │             │   │                                           
                                          │ │    Image    │   │                                           
                                          │ │             │   │                                           
                                          │ ├─────────────┤   │                                           
                                          │ │  Link text  │   │                                           
                                          │ └─────────────┘   │                                           
                                          │                   │                                           
                                          └───────────────────┘                                           
                                                                                                          
                                          ┌───────────────────┐                                           
                                          │       Blocks      │  We can repeat multiple                   
                                          │                   │   different groups of                     
                                          │ ┌─────────────┐   │   fields, one or more                     
                                          │ │    Title    │   │          times                            
                                          │ ├─────────────┤   │                                           
                                          │ │             │   │                                           
                                          │ │    Image    │   │                                           
                                          │ │  ┌──────────┴──┐│                                           
                                          │ ├──┤   Heading   ││                                           
                                          │ │  ├─────────────┤│                                           
                                          │ └──┤             ││                                           
                                          │    │    HTML     ││                                           
                                          │    │             ││                                           
                                          │    ├─────────────┤│                                           
                                          │    │   Button    ││                                           
                                          │    └─────────────┘│                                           
                                          └───────────────────┘                                           
                                                                                                          
                                                                                                          
                                                                                                          
                                                                                                          
                                                                                                          
    This is how it works in Bolt 4:                                                                       
                                                                                                          
     - First we take our intended                                                                         
   groups of fields, and call them a                                                                      
                 Set:                    ┏━━━━━━━━━━━━━━━┓                   ┏━━━━━━━━━━━━━━━┓            
                                         ┃┌─────────────┐┃                   ┃┌─────────────┐┃            
                                         ┃│    Title    │┃  Set "A"          ┃│   Heading   │┃    Set "B" 
                                         ┃├─────────────┤┃                   ┃├─────────────┤┃            
                                         ┃│             │┃                   ┃│             │┃            
                                         ┃│    Image    │┃                   ┃│    HTML     │┃            
                                         ┃│             │┃                   ┃│             │┃            
                                         ┃├─────────────┤┃                   ┃├─────────────┤┃            
                                         ┃│  Link text  │┃                   ┃│   Button    │┃            
                                         ┃└─────────────┘┃                   ┃└─────────────┘┃            
                                         ┗━━━━━━━━━━━━━━━┛                   ┗━━━━━━━━━━━━━━━┛            
                                                                                                          
                                                                                                          
                                                                                                          
                                                                                                          
                                                                                                          
                                        ┌───────────────────────────┐                                     
    And then we make a Collection, by   │      Collection           │                                     
      configuring it to use these:      │                           │                                     
                                        │ ┏━━━━━━━━━━━━━━━┓         │                                     
                                        │ ┃┌─────────────┐┃         │                                     
                                        │ ┃│    Title    │┃         │                                     
                                        │ ┃├─────────────┤┃         │                                     
                                        │ ┃│    ┏━━━━━━━━┻┻━━━━━┓   │  We can make a collection, where we 
                                        │ ┃│    ┃┌─────────────┐┃   │  define a bunch of fields to choose 
                                        │ ┃│    ┃│   Heading   │┃   │  from. The editor can then mix and  
                                        │ ┃├────┫├─────────────┤┃   │  match these to build the content as
                                        │ ┃│  Li┃│             │┃   │  they desire.                       
                                        │ ┃└────┫│    HTML     │┃   │                                     
                                        │ ┗━━━━━┫│             │┃   │  The flexibility comes from being   
                                        │       ┃├─────────────┤┃   │  able to pick all fields:           
                                        │       ┃│   Button    │┃   │                                     
                                        │       ┃└─────────────┘┃   │  - single regular fields (so it     
                                        │       ┗━━━━━━━━━━━━━━━┛   │  behaves similar to old "repeaters")
                                        │  ┌─────────────┐          │  - Sets of fields (so it behaves    
                                        │  │  Link text  │          │  similar to old Blocks)             
                                        │  └─────────────┘          │  - or a combination, for extra      
                                        │           ┌─────────────┐ │  versatility                        
                                        │           │             │ │                                     
                                        │           │    Image    │ │                                     
                                        │  ┌────────┤             │ │                                     
                                        │  │        └────┬────────┘ │                                     
                                        │  │    HTML     │          │                                     
                                        │  │             │          │                                     
                                        │  └─────────────┘          │                                     
                                        └───────────────────────────┘                                     
```

eduardomart:
I understand the diagram but I am still not sure I understand why we need Sets.

In v3 we had something like this (from the [docs](https://docs.bolt.cm/3.7/fields/block#basic-configuration))

```
contentblocks:
            type: block
            label: Content Blocks
            group: Blocks
            fields:
                imagesection:
                    label: Image with description
                    fields:
                        heading:
                            type: text
                        image:
                            type: image
                        description:
                            type: html
                            height: 70px
                textsection:
                    label: Paragraph
                    fields:
                        heading:
                            type: text
                        content:
                            type: html
                            height: 100px
```

We could easily add single fields in there as well without requiring the extra field type, for example:

```
contentblocks:
    type: block
    label: Content Blocks
    group: Blocks
    fields:
        description:
           label: single text field # <-------------
            fields:
                desc:
                    type: text
        singleimage:
           label: single image field # <-------------
            fields:
                img:
                    type: image
        imagesection:
            label: Image with description
            fields:
            heading:
                type: text
            image:
                type: image
            description:
                type: html
                height: 70px
        textsection:
            label: Paragraph
            fields:
            heading:
                type: text
            content:
                type: html
                height: 100px
```

bobdenotter:
@eduardomart 

> We could easily add single fields in there as well without requiring the extra field type, for example:

Well, yes.. it was still confusing to people, though. 

What we have with v4 now might seem like it's making it more complex, it makes much sense under the hood. Inside a `collection`, fields are stored as _actual_ fields. As such, we _required_ a mechanism to group them in order to be able to add them to a collection in one piece. So, that's how sets originated. And since we have them, might as well expose them to users. 

Even if they're not very useful for the majority of users, it might still be convenient to have them for some specific usecases.

eduardomart:
Ok, that makes sense. Thanks @bobdenotter.

PR #1027, opened 2020-04-01
Update Relationships Documentation

technicallyerik:
* Clarified the contenttypes.yml example, and how relations are associated
* Add information about the "required" option
* Update templating example to use the "related" twig filter
* Remove the paragraph about the "relation" property on the content (this no longer exists)
* Remove the paragraph about the "related()" twig function (this no longer exists)


bobdenotter (approved):
Thanks, @technicallyerik 👍

PR #1028, opened 2020-04-01
Taxonomy Documentation Updates

technicallyerik:
* Clarify how "slug" and "name" are used
* Updated the default taxonomy config to match the latest that is generated for a new site
* Add the "required" option to the properties listing
* Fix the twig example to list the taxonomy on a record
* Add examples on how you can display all categories or tags in a sidebar


bobdenotter (approved):
Thanks again! :shipit:

PR #1032, opened 2020-04-10
add missing doc about using json manifest with asset function

dadaxr:
that doc was previously created and merged in #944
was also merged with 3.6 branch with #965 
but was never merged into 3.X and 3.7 branch...
this is the PR for the 3.x branch !


bobdenotter (approved):
Thanks for both of these! :+1:

PR #1100, opened 2020-09-16
Fixes syntax for link to tag

madc:
This is a minor PR that introduces the new link syntax via filter.


I-Valchev (approved):
Good catch! Thanks @madc

PR #1104, opened 2020-09-24
Title number.md

evertalbers:
You get the idea - I was a bit confused  of the use of both 'float' and 'number'. Maybe php'ers will be looking for 'float' field though - so maybe leave both in the title. Feel free to change this into what you think is best.


bobdenotter (approved):
Thanks, @evertalbers 👍

PR #1107, opened 2020-09-25
fix home link location

dadaxr:
now redirecting to boltcms.io instead of bolt.cm :)


I-Valchev (approved):
Good spot, thanks @dadaxr

PR #1120, opened 2020-11-10
Update singleton-contenttype.md

simonmnt:
The geolocation type no longer exists.


bobdenotter (approved):
You're correct! Thanks, @simonmnt

PR #1122, opened 2020-11-13
Update content-paging.md

simonmnt:
Bad path.


bobdenotter (approved):
Good catch, thank you @simonmnt 👍

PR #1125, opened 2020-11-18
Update introduction.md

prdurand:
as a newbie in composer, trying "delete" didn't work for me, but "remove" yes, so I propose the change


I-Valchev (approved):
Nice catch, thanks @prdurand! 👍

PR #1127, opened 2020-11-26
Fix broken link

Chi-teck:


I-Valchev (approved):
Good catch, thanks @Chi-teck 👍

PR #1132, opened 2020-12-08
Error in homepage example

peterboorsma:
The default content to use for the homepage, and the template to render it
with. This can either be a singleton like `homepage`, a specific record (like
`pages/1`)

The example says "page/1" but it should be "pages/1"


bobdenotter:
Thanks! 👍

bobdenotter (approved):

PR #1136, opened 2020-12-17
Nog should be not.

sitedyno:


I-Valchev (approved):
"Nog" 🤦‍♂️

Good catch, thank you @sitedyno .

PR #1140, opened 2021-02-06
Add nginx config restriction exception for letsencrypt

lucaslcode:
Let's Encrypt's http challenge uses `.well-known/acme-challenge/xxxx` for verification

Denying access to everything starting with a dot was causing my certificate renewal to fail.


bobdenotter (approved):
Nice catch! Thank you, @lucaslcode 👍

PR #1143, opened 2021-02-17
Modify DATABASE_URL for MySQL

tambait:


bobdenotter:
Hi @tambait, 

Can you explain a bit why this is a beneficial change? I've only ever bumped into passwords with special characters needing escaping, but not the entire DSN.

tambait:
Hi @bobdenotter,
my mistake, I was getting MySQL error during installation using original syntax (while no escaping or escaping whole DSN worked). But I checked it now again it was not related to the syntax used but my server configuration.

bobdenotter:
@tambait Thanks for the feedback, and glad you managed to fix it!

PR #1148, opened 2021-03-03
Delete reference to has_sortorder

evertalbers:
I think `has_sortorder` was something from Bolt 3 which has not been implemented in Bolt 4 (yet?). 
FYI there's also 2 references in https://docs.bolt.cm/4.0/contenttypes/taxonomies


bobdenotter:
We might want to bring it back someday (if there's demand), but for now let's remove it! 

Thanks, @evertalbers 👍

bobdenotter (approved):

evertalbers:
It is not really necessary when there's other ways to order results. What I miss more is grouping contenttypes in the backend by a visible 'grouping' taxonomy - but I vaguely recall you mentioning that in the online meeting a few weeks back.

PR #1150, opened 2021-03-22
Added example fetching content with equal taxonomy values

peterboorsma:
Added "Show records with the same taxonomy value".


I-Valchev (approved):
The more, the better!

Thanks @peterboorsma, merging it in right away!

PR #1151, opened 2021-04-17
Update fetching of records in select field

DorienD:
Update the fetching of records when populating the select field with multiple ContentTypes


bobdenotter (approved):
Thanks, @DorienD 👍

PR #1166, opened 2021-06-28
Calculating in Twig with integers

evertalbers:
Following Ivo's tips on https://boltcms.slack.com/archives/C094GL7ME/p1624866867086300

Technically the same could be added to the float field, but float-users will probably see the mention at the integer when scrolling past the options.

### DIFF
--- a/docs/manual/fieldtypes.md
+++ b/docs/manual/fieldtypes.md
@@ -104,6 +104,11 @@ true if the checkbox was checked, and 0 for false if it wasn’t checked.
 A field to store whole numbers. The value must be between -2147483648 and
 +2147483647. Often used for storing things like prices of products.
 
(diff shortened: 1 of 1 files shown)


bobdenotter (approved):
Thanks, @evertalbers 👍

evertalbers:
Oh darn, I think I added this at the wrong location. Looks like this should have gone into https://docs.bolt.cm/4.0/fields/number and not in the end user manual.

bobdenotter (approved):

bobdenotter (changes requested):
Ah, right.. Should we close this one, then?

evertalbers:
Yes, please close it

bobdenotter:
Done. ;-)

Issue #1168, opened 2021-06-29
TODO's for Bolt 5 documentation:


bobdenotter:
- [ ] Write about how to deploy a Bolt site to production (installation/deployment.md) @bobdenotter 
 - [x] Write about how to create custom Twig tags and Filters (extensions/twig.md) @I-Valchev 
 - [x] Write some stuff here adding pages in the backend, that kinda look like they fit in with the overall backend (extensions/pages.md) @bobdenotter or @I-Valchev 
 - [ ] Menus (extensions/menu.md) @bobdenotter 
 - [x] Entities (extensions/entities.md) @I-Valchev 
 - [x] Creating content programatically (extensions/creating_content.md) @I-Valchev 
 - [x] Using Composer packages (extensions/composer.md) @I-Valchev 
 - [x] Fetching content (extensions/fetching_content.md) @I-Valchev 
 - [ ] Logging (extensions/log.md) @bobdenotter 
 - [x] Events (extensions/event.md) @I-Valchev 
 - [ ] Write some stuff here about Autowiring and how to fetch services using `$this->services` (extensions/services.md) @I-Valchev or @bobdenotter 
 - [ ] Request / Response (extensions/request.md) @bobdenotter 
 - [ ] Global Config (extensions/global_config.md) @bobdenotter 
 - [ ] Making a autoloadable Configuration object (and add it to the reference object) @bobdenotter 
 - [ ] Docs on upgrading (upgrading/updating.md) @bobdenotter 
 - [ ] Multilingual websites (howto/building-multilingual-websites.md) has a bunch of todos @bobdenotter 
 - [ ] Setting up email, and testing it: Weird DSN format, use of https://mailcatcher.me/ or https://github.com/mailhog/MailHog @bobdenotter 
 - [x] Using a callback for `type: select` fields @I-Valchev 
 - [ ] Better description for `{{ menu() }}` with parameters @bobdenotter

bobdenotter:
Another page that needs to be added: 

Setting up email, and testing it: Weird DSN format, use of https://mailcatcher.me/ or https://github.com/mailhog/MailHog

I-Valchev:
Ah yes that'd be great!

Also, should we keep that PR #1169 open while filling in the pages? Seems like a nice way to keep track.

bobdenotter:
I think we should merge that one (like the incremental updates for the twig methods), and turn this one into a checklist. That'll make it easier to work on it piece by piece.

PR #1169, opened 2021-06-29
Working on docs for v5, mostly Extensions

bobdenotter:
Working on the issues in #1168


I-Valchev (approved):
As discussed in #1168, let's merge this as-is and fill it in gradually.

PR #1176, opened 2021-07-10
Unhide <link rel="shortcut icon">

evertalbers:


bobdenotter:
Thanks! while your fix undeniably makes it better (as in: syntactically correct), i wonder if this is correct. Did we port that functionality to v4? Is it even desirable that Bolt injects magical links to shortcut icons? 🤔

evertalbers:
Well, actually I think not so you may want to delete lines 90-98 altogether. To be honest I prefer to manually add the favicon links to a Twig template, as that's easier to debug.

PR #1180, opened 2021-07-23
Disclaimerish note about inline svg's

evertalbers:
Tried to keep the text as generic as possible, to be on the safe side.


bobdenotter (approved):
Sounds good! 👍

PR #1189, opened 2021-08-23
Change integer/float to number & mode integer/float

evertalbers:
Referring to https://boltcms.slack.com/archives/C094GL7ME/p1629719308362000


I-Valchev (approved):
Nice catch, thanks @evertalbers 👍

PR #1193, opened 2021-09-06
Fix typo

garrettboone:
Now version 5


I-Valchev (approved):
👍 Nice catch, thanks @garrettboone

garrettboone:
@I-Valchev On the same page, I wasn't sure if `With the stable release of Bolt 4,` should also be considered...I left it because it seemed like an event reference.

PR #1198, opened 2021-09-22
Add title block to example to prevent errors

nestordedios:
If the title block is not added and a developer uses this example they will see an error saying that the title block does not exist.

This PR prevents other developers bumping into this.


bobdenotter (approved):
Thanks, @nestordedios

PR #1200, opened 2021-10-06
Updating the updating instructions

evertalbers:
Assuming `5` is better than using `%%VERSION%%`


I-Valchev (approved):
👍 Looks good, thank you @evertalbers !

PR #1201, opened 2021-10-07
Remove outdated localedatetime documentation

bobvandevijver:
Was already deprecated in 4.0 (https://github.com/bolt/core/pull/1612), so I guess this should be merged in that branch already. Otherwise feel free to update it to be 5.0 only.

### DIFF
--- a/docs/localization/locales.md
+++ b/docs/localization/locales.md
@@ -60,7 +60,6 @@ sv_SE    Swedish              Sweden                           sv   SE
 A much longer list of possible options can be found here:
 [List of Locales, languages and countrycodes](https://github.com/bobdenotter/locales/blob/master/locales_list.txt).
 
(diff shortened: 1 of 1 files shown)


I-Valchev:
Hi @bobvandevijver , thanks for this PR! You're right that it still works in Bolt 4 (although deprecated at some point) and the new function works in 5.0.

So, do you think we ought to update the docs with your contribution for 5.0 only?

bobvandevijver:
In my opinion it should be 4.0, as the `localedatetime` filter actually isn't described anymore on the linked filters page.

I-Valchev:
Sure, agreed! Thanks again @bobvandevijver !

bobvandevijver:
@I-Valchev Are you also merging this into the 5.0 branch, or would you like a separate PR for that?

I-Valchev:
I've created #1206 as there's a few other things we can bring into 5.0 :-)

PR #1202, opened 2021-10-09
Improve the Nginx configuration

tunggnu:
- Where capturing group is not used, `?:` is redundant. Therefore, I removed those `?:`. (Ref: https://www.pcre.org/original/doc/html/pcresyntax.html#SEC13)
- `$is_args$args` is more accurate than `?$query_string` because `$args` is same as `$query_string` and if `$args` is empty then `$is_args` is empty too.
- Replace `return 403;` with `deny all;` for consistency. 
- `location = /(?:favicon.ico|robots.txt)` is an exact match but `/(?:favicon.ico|robots.txt)` is a regex pattern, therefore it does not work as expected.
- The two location blocks for Apache and SQLite are redundant because there is already a block right before them to block access to "hidden" files.

### DIFF
--- a/docs/installation/webserver/nginx.md
+++ b/docs/installation/webserver/nginx.md
@@ -82,7 +82,7 @@ on a host.
 ```nginx
 # Default prefix match fallback, as all URIs begin with /
 location / {
-    try_files                     $uri $uri/ /index.php?$query_string;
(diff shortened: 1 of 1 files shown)


bobdenotter:
Excellent additions! Thank you, @tunggnu 👍

PR #1203, opened 2021-10-12
Update taxonomy config access prefix

nestordedios:
The prefix to access Taxonomy config (taxonomy.yaml) seems to be `taxonomies` instead of `taxonomy`.


bobdenotter:
Thanks, @nestordedios 👍

PR #1204, opened 2021-10-12
Update maintenance-mode.md

evertalbers:
Please notice that I could not directly find a default template file in the Bolt repo, but if there is one there, my text should be slightly rewritten.


bobdenotter (approved):
Looks good to me, @evertalbers 👍

PR #1207, opened 2021-10-27
Fix misspelled on production env tips

zemian:


I-Valchev (approved):
Nice catch, thanks @zemian

PR #1209, opened 2021-10-29
Describe lazy loading for |showimage

evertalbers:
and update cropping info, since only 'contain' mode is supported.


bobdenotter:
Thanks, @evertalbers 👍

PR #1210, opened 2021-10-30
Fix has example

nicodeboer:


bobdenotter (approved):
Good catch! Thanks, @nicodeboer 👍

PR #1213, opened 2021-11-01
Added note for Bolt3 recordsperpage to records_per_page

michaelborn:
IMO, this should be mentioned on the "Migrating from Bolt v3.7 to 5.0" docs.

### DIFF
--- a/docs/upgrading/upgrading-from-30.md
+++ b/docs/upgrading/upgrading-from-30.md
@@ -137,12 +137,26 @@ integerfield:
     mode: integer
 ```
 
-To migrate the data of an **Integer** field the new field definition will look like:
+To migrate the data of a **Float** field the new field definition will look like:
 ```yaml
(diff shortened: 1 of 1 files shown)


michaelborn:
**Note:** This Bolt 4.0 docs page also has an outdated example using `recordsperpage`, but I can't find the `4.0` tag/branch in Github to submit a quick PR.

https://docs.boltcms.io/4.0/contenttypes/intro#defining-contenttypes

bobdenotter:
Good catch, thank you! 👍 

> I can't find the 4.0 tag/branch in Github to submit a quick PR.

Ugh.. If we merge down branches, there's this "helpful" button to delete a branch after merging. Often this is what you want, but in this case it causes the deletion of a branch we should keep.. It's back now: https://github.com/bolt/docs/tree/4.0

PR #1229, opened 2022-02-25
Explain how to customize password reset emails

vlad-ghita:


bobdenotter (approved):
Awesome, thank you @vlad-ghita 👍

PR #1234, opened 2022-03-02
warn about reserved assets package names

dadaxr:
Is could be tempting (as I did) to use the "theme" assets package name (instead of "my-site). But the config is never effective because it's overriden by bolt "TwigAwareController"  which defined some twig packages (used internally by the {{ asset() }} function), see TwigAwareController line 206 for details...

reserved packages names are : "bolt" / "theme" / "public" and "files"


bobdenotter (approved):
Good point, thanks @dadaxr 👍

PR #1237, opened 2022-03-21
Update content-fetching.md

Joossensei:
Update the query because the old doesnt work.

Edit: Only works in some cases see [#3130](https://github.com/bolt/core/issues/3130)


mcdennem:
👍

PR #1240, opened 2022-03-29
Fix name of "extensions:configure" docs page

michaelborn:
... plus add a bit more explanation about what this command _actually_ does.


mcdennem:
👍

PR #1248, opened 2022-05-24
Change namespace to App\Color

shadowbytefox:
Use App\Color as namespace to avoid confusion.


shadowbytefox:
Bolt\Color does not work, only App\Color seems to be allowed.

bobdenotter (approved):
Thanks, @shadowbytefox 👍

PR #1249, opened 2022-05-27
Fix broken pager example

michaelborn:
Resolves this error I get when using the example code:

```
Positional arguments cannot be used after named arguments for function "pager".
```


bobdenotter (approved):
Thanks, @michaelborn 👍

Issue #1239, opened 2022-03-23
Non existing setting 'attrib' in nodes example


peterboorsma:
This page https://docs.boltcms.io/5.0/contenttypes/intro#advanced-yaml-repeated-nodes uses `attrib: title `on the image field in the example. As far as I know `attrib` is not used anymore. The only option is `alt: true`. 

Please let me know if I'm right or wrong so I can edit the page correctly. :)

PR #1250, opened 2022-05-29
Change image options attrib: title to alt: true

shadowbytefox:
attrib is not found among the options of the type image field.
deprecated? alt: true is a good alternative here to maintain consistency

Issue Nr:  #1239


bobdenotter (approved):
Thanks, @shadowbytefox 

Apologies for the long time this has been in the queue

PR #1252, opened 2022-06-01
Suggestion for different twig example code for file fields

evertalbers:


bobdenotter (approved):
Nice catch. Thanks, @evertalbers 👍

PR #1253, opened 2022-06-10
Add `limit` as `latest` directive doesn't take numbers

xiaohutai:
See: https://github.com/bolt/core/issues/3220


bobdenotter (approved):
Nice catch, thanks!

PR #1255, opened 2022-06-22
Document previous and next classes for pagination component

nestordedios:


nestordedios:
This documents new feature https://github.com/bolt/core/pull/3255

PR #1259, opened 2022-08-03
Update the coding standard from PSR-2 to PSR-12

abdellahrk:


bobdenotter:
It looks like this PR overwrites some changes. We were perhaps editing at the same time. Going to close this one, because it should be OK now: https://docs.boltcms.io/5.0/other/code-quality#code-standard

Thanks, regardless. 🙏

PR #1260, opened 2022-08-04
remove documentation for "Log Management"

simongroenewolt:
Remove documentation for "Log Management" as there are no console utilities for that anymore since Bolt 4. It looks like this was just documentation for tools provided by Silex.

I saw a question by someone on Slack about the "Log Management" commands. I don't think it makes sense to re-introduce these commands, so it is better to remove the associated doc and prevent further confusion in the future.


bobdenotter (approved):
Thanks, @simongroenewolt 👍

PR #1267, opened 2022-10-17
Remove deprecated instructions about template fields

caillioux:
Bolt had in version 3.x a feature allowing to add template fields.

According to @bobdenotter this feature does not exist anymore, thus I propose removing the information about specific template fields in the documentation.


bobdenotter (approved):
Thanks for the clarification, @caillioux 👍

PR #1286, opened 2025-08-21
Update dead link

evertalbers:


docs/twig-components/method/setcontent.md
```diff
@@ -4,7 +4,7 @@
 records.
 
 Much, much more information about the `setcontent` tag, together with additional query arguments, pagination, sorting 
-and other options you can find in the chapter about [Fetching content](https://docs.bolt.cm/4.0/templating/content-fetching).
+and other options you can find in the chapter about [Fetching content](../templating/content-fetching).
```
bobvandevijver:
```suggestion
and other options you can find in the chapter about [Fetching content](../../templating/content-fetching).
```

It think it should be this, right?

evertalbers: You're probably right, I'm not good at counting ../'s

bobvandevijver:
Thanks!

